Issue #3, opened 2023-03-09
Add a linter


parkerduckworth:
We curerntly do not have a linter for the JS code. I don't have any preferences so please use the most common one :)

This should be added:

In the CI
In [precommit](https://pre-commit.com/)

Issue #2, opened 2023-03-09
Add a formatter


parkerduckworth:
We curerntly do not have a formatter for the JS code. I don't have any preferences so please use the most common one :)

This should be added:

In the CI
In [precommit](https://pre-commit.com/)

PR #4, opened 2023-03-09
feat: linter and formatter

mloenow:
Closes https://github.com/weaviate/typescript-client/issues/2
Closes https://github.com/weaviate/typescript-client/issues/3

Adds `eslint` and `prettier` with TypeScript plugins. All configuration is adopted from https://github.com/stripe/stripe-node with some smaller modifications (e.g. changing errors to warnings, personal format preferences). Fixed all issues given by the linter, but there are still 57 warnings left (besides use of `any`).

This pull request also moves most of the source code into `src/` for better separation of project artifacts (https://github.com/weaviate/typescript-client/commit/41e2bc9ab963430d8049798d914d5a4d71f9feae).

@parkerduckworth (or @dirkkul, as you added the original issues) - can you assist with the CI and precommit integration? I can add it if I get some pointers 🙂 

### TODO

- [x] CI
- [x] Pre-commit
- [x] Fix tests


mloenow:
The linter complains quite a bit more after the recreation, so I think something is off. Will look at it again tomorrow (midnight here).

mloenow:
Ok, found the issue - only warnings now, like before 😄

dirkkul:
HI @mloenow 

thanks a bunch for your contribution!

Here is a pointer for [github actions with eslint and prettier](https://profy.dev/article/continuous-integration-with-github-actions/). Just add a new linting workflow file :)

For precommit, have a [look here for prettier](https://prettier.io/docs/en/precommit.html#option-3-pre-commithttpsgithubcompre-commitpre-commit). I can't find anything easy for eslint, here is a bit [more complex one](https://justinnoel.dev/2019/07/13/configuring-precommit-hooks-for-eslint-and-prettier/). I am not familiar with TS, so I don't have a feeling how complex that is. Just ignore this one if it is too annoying.

For pre-commits in general: It allows you to run linters/formatters automatically at each commit, avoiding annoying CI failures. SImply install on your system (`brew install pre-commit` / `apt install pre-commit`) go to the repo folder and do `pre-commit install` and it is enabled :)

dirkkul:
I think we should merge this quickly before we get any conflicts. we can always do pre-commit/CI later

mloenow:
OK. I will set it ready for review and you guys can get it merged if all is good.

.npmignore
```diff
@@ -1,16 +1,11 @@
...
+# ignoring everything, then add back the files we want to be included in the
+# final NPM package.
+*
+
+# And these are the files that are allowed.
+!/LICENSE
+!/README.md
+!/package.json
+!/dist/**/*
+!/types/**/*
```
parkerduckworth, line 11: Can you please add back all the directories and files which were being ignored

.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
+  "arrowParens": "always",
+  "trailingComma": "es5",
+  "bracketSpacing": true,
+  "singleQuote": true
+}
```
parkerduckworth, line 6: Please add newline at the end of this file

package.json
```diff
@@ -3,19 +3,23 @@
...
+    "/dist/index.js",
+    "/dist/index.js.map",
+    "/types/**/*.d.ts"
   ],
   "scripts": {
-    "test": " tsc -noEmit -p tsconfig-test.json && jest --useStderr --runInBand --detectOpenHandles --forceExit",
-    "build": "tsc --emitDeclarationOnly && ./build.js",
-    "prepack": "npm run build"
+    "test": "tsc -noEmit -p tsconfig-test.json && jest --useStderr --runInBand --detectOpenHandles --forceExit",
+    "build": "npm run lint && tsc && ./build.js",
```
parkerduckworth, line 17: Why was the `--emitDeclarationOnly` removed?

package.json
```diff
@@ -44,11 +48,17 @@
...
     "babel-jest": "^29.4.3",
     "esbuild": "^0.17.10",
+    "eslint": "^8.35.0",
+    "eslint-config-prettier": "^8.7.0",
+    "eslint-plugin-prettier": "^4.2.1",
     "jest": "^29.4.3",
-    "prettier": "^2.1.1",
+    "prettier": "^2.8.4",
     "ts-jest": "^29.0.5",
+    "ts-node": "^10.9.1",
```
parkerduckworth, line 61: What is the `ts-node` package needed for?

mloenow:
Formatting + linting CI integration added now, but some tests are failing it seems.

src/data/path.ts
```diff
@@ -0,0 +1,206 @@
...
+    consistencyLevel: string | undefined
+  ): Promise<string> {
+    return this.build({ id, className, consistencyLevel }, [
+      this.addClassNameDeprecatedCheck,
+      this.addId,
+      this.addQueryParams,
+    ]);
+  }
+
+  build(params: any, modifiers: any): Promise<string> {
```
parkerduckworth, line 82: Curious about the inconsistency of newlines in between this class' methods.

For example there is not a newline in between all the above methods up to the constructor, but starting here at `build` there is one, and below there is no newline between the remaining methods

src/index.ts
```diff
@@ -0,0 +1,100 @@
...
+    // check if the scheme is set
+    if (!params.scheme) throw new Error('Missing `scheme` parameter');
+
+    // check if headers are set
+    if (!params.headers) params.headers = {};
+
+    const conn = new Connection(params);
+    const dbVersionProvider = initDbVersionProvider(conn);
+    const dbVersionSupport = new DbVersionSupport(dbVersionProvider);
+    const test = [1, 2, 3];
```
parkerduckworth, line 61: What is this for?

parkerduckworth (changes requested):
I can't comment on it because there were no changes to the file, but please put `src/tools/prepare_release.sh` back into the project root

parkerduckworth:
As far as the failing tests go, can you see the [test logs](https://github.com/weaviate/typescript-client/actions/runs/4386254279/jobs/7680081974?

The failures are due to some of the expected responses from the server coming back slightly different. 

For example, some tests are just checking for a status code, but the server is returning the full error object (with the expected status code), etc

.npmignore
mloenow: I can add it back, but this file now works like a whitelist instead, so you only have to add what you actually want published (see the * in the beginning for blacklist of everything - then after comes the exceptions). Personally I think that's cleaner/safer.

Let me know if I should still revert.

package.json
mloenow: No preference - I think it was just because merging changes from the two repos.

tsconfig.json
```diff
@@ -4,16 +4,25 @@
     "strict": true,
     "preserveConstEnums": true,
     "noEmit": false,
+    "emitDeclarationOnly": true,
```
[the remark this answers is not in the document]
mloenow: No preference - I think it was just because merging changes from the two repos.

Re-adding as CLI flag.

src/index.ts
mloenow: I don't think this is from my changes? 🤔 It's not something I added manually

src/data/path.ts
mloenow: I don't think there's a linting rule enforcing this currently, but we should be able to add it. Should I?

src/index.ts
parkerduckworth: Weird, its [not in the current HEAD](https://github.com/weaviate/typescript-client/blob/main/index.ts#L50-L54) either way if you don't mind removing it 🙂

mloenow: Ah, you meant just the last line - and now I recall it - was testing linting/formatting 😆

package.json
mloenow: Will verify if it's needed or not.

mloenow:
The pre-commit lint check now won't let you commit due to excessive warnings, but this can be ignored with `--no-verify`.

I will look and the remaining things in a few hours.

dirkkul:
> The pre-commit lint check now won't let you commit due to excessive warnings, but this can be ignored with --no-verify.

Then let's disable it for now :) There is no point in a checker that is failing by default
Just rename the file then we can enable it once the warnings are reduced?

mloenow:
> Just rename the file then we can enable it once the warnings are reduced?

It's part of `package.json` - but can easily be re-added 🙂

mloenow:
The `module.exports = app;` export recently added in `src/index.ts` might not be the best way to go about it for CommonJS support, as `esbuild` now gives a warning:

```
▲ [WARNING] The CommonJS "module" variable is treated as a global variable in an ECMAScript module and may not work as expected [commonjs-variable-in-esm]

    src/index.ts:98:0:
      98 │ module.exports = app;
         ╵ ~~~~~~

  This file is considered to be an ECMAScript module because of the "export" keyword here:

    src/index.ts:100:0:
      100 │ export { AuthUserPasswordCredentials, AuthAccessTokenCredentials, Operator };
          ╵ ~~~~~~
```

I don't have experience with `esbuild`, so I don't know how this can be improved.

package.json
mloenow: Don't believe it is. Bye 👋

mloenow:
> As far as the failing tests go, can you see the [test logs](https://github.com/weaviate/typescript-client/actions/runs/4386254279/jobs/7680081974)?
> 
> The failures are due to some of the expected responses from the server coming back slightly different.
> 
> For example, some tests are just checking for a status code, but the server is returning the full error object (with the expected status code), etc

Tests have been fixed. The after lint fixes Promise rejection are using proper Error objects instead of just messages. I think it's consistent now.

@parkerduckworth @dirkkul could you take another look? 🍻

mloenow:
I agree to the CLA 👍

parkerduckworth:
@mloenow nice job, tests are passing :)

WRT to the `module.exports`, the problem is that the `index.ts` uses a default export. Without `module.exports`, importing the client from a compiled TS project is broken, because the TS compiler does not add in the extra `default` qualifier in when calling `weaviate.client({...})`, resulting in `weaviate.client is not a function`.  

The correct usage in this case would be `weaviate.default.client({...})`, or the import would need to be `const weaviate = require('weaviate-ts-client').default`.

Maybe there is a way to get `tsc` to include this in the output .js files, but if so I am not aware of it.

---

So, aside from just leaving `module.exports` in this library's index.ts, which is arguably bad style by mixing CommonJS and ESM modules in the same file, I have found a workaround: not using any default exports.

Check out [this branch](https://github.com/weaviate/typescript-client/compare/no-default-export) to see the changes.

I'm curious to know what you think about this change.

The one thing I am unsure about is that by going this route, all TS projects wanting to use the client would have to import the client like this:

```js
import pkg from 'weaviate-ts-client';
const { weaviate } = pkg;
```

Which seems like prioritizing CommonJS over ESM. Not sure if that is a good or bad thing.

parkerduckworth (approved):
We can go ahead and merge this. The precommit, other linter errors, etc can be added separately. Thanks again for the contribution @mloenow !

PR #8, opened 2023-03-14
Support embedded weaviate

parkerduckworth:
Embeds a Weaviate instance in the client, so there is no need to run the server elsewhere, in preparation for TS notebook support.

Usage:
```typescript
import weaviate from 'weaviate-ts-client';

const client = weaviate.client({
  scheme: 'http',
  host: 'localhost:9898',
  embedded: new weaviate.EmbeddedOptions({
    port: 9898,
  }),
});

await client.embedded?.start();

client.misc
  .metaGetter()
  .do()
  .then((res: any) => console.log(`res: ${JSON.stringify(res)}`));

client.embedded?.stop();
```

Note:
> Embedded DB can currently only be run on a Linux OS/docker container. MacOS will be supported in the future


src/embedded/index.ts
```diff
@@ -0,0 +1,301 @@
...
+  host: string;
+  port: number;
+  clusterHostname: string;
+  version: string;
+  env: NodeJS.ProcessEnv;
+
+  constructor(cfg?: EmbeddedOptionsConfig) {
+    this.clusterHostname = 'embedded';
+    this.host = (cfg && cfg.host) || '127.0.0.1';
+    this.port = (cfg && cfg.port) || 6666;
```
antas-marcin, line 34: should be here some random port? or we agreed to make it default to `6666`?

src/embedded/index.ts
```diff
@@ -0,0 +1,301 @@
...
+    this.version = this.setVersion(cfg);
+    this.binaryPath = this.setBinaryPath(cfg);
+    this.persistenceDataPath = this.setPersistenceDataPath();
+    this.env = this.setEnv(cfg);
```
antas-marcin, line 38: instead of `setXXX` wouldn't it better to prefix those methods with `getXXX` or `parseXXX`. For me setters don't return any value they just set the value inside the method body

src/embedded/index.ts
parkerduckworth: Unfortunately getting a random available port in nodes requires an async function, which would mean creating an `EmbeddedOptions` instance needs to be `await`'ed.

This is the same default port which the python client embedded db uses.

antas-marcin: ok, let's leave it then

src/embedded/index.ts
parkerduckworth: agreed, changing it

antas-marcin (approved):

PR #33, opened 2023-03-31
Fix: Blocking background OIDC token refresh

### LINKED ISSUE #30: Script not exiting when using `AuthUserPasswordCredentials`
When running the exact same script, instantiating the client with `AuthUserPasswordCredentials` leads to the script not exiting upon completion. For example:

```
import weaviate, { WeaviateClient } from 'weaviate-ts-client';

const client: WeaviateClient = weaviate.client({
  scheme: "https",
  host: "<wcs instance w OIDC>",
  authClientSecret: new weaviate.AuthUserPasswordCredentials({
    username: "<email>",
    password: "<password>",
  })
});

client
  .schema
  .getter()
  .do()
  .then((res: any) => {
    console.log(res);
  })
  .catch((err: Error) => {
    console.error(err)
  });
```

Causes it to hang at completion (running with `ts-node`), whereas authenticating with the API key will work just fine:

```
import weaviate, { WeaviateClient } from 'weaviate-ts-client';

const client: WeaviateClient = weaviate.client({
  scheme: "https",
  host: "<the same wcs instance w OIDC>",
  headers: {"Authorization": "Bearer <API key>"}
});

client
  .schema
  .getter()
  .do()
  .then((res: any) => {
    console.log(res);
  })
  .catch((err: Error) => {
    console.error(err)
  });
```

Note: This behaviour (hanging) does not occur when using a WCS instance without authentication turned on, either.

parkerduckworth:
A `setInterval` is used to run a background OIDC token refresh, but ends up blocking a script from exiting due to the interval never being cleared. 

This PR adds two changes to surmount this issue:

1. An explicit `client.oidcAuth?.stopTokenRefresh()` method which should be used whenever a script is expected to exit, or it is otherwise determined that token refresh is no longer needed
2. A `silentRefresh?: boolean` argument to each of the OIDC credential inputs, including `AuthClientCredentials`, `AuthUserPasswordCredentials`, and `AuthAccessTokenCredentials`. This currently defaults to `true` if not set, as this is likely the expected behavior for most users.

Closes #30

### DIFF
--- a/src/connection/auth.ts
+++ b/src/connection/auth.ts
@@ -6,17 +6,22 @@ interface AuthenticatorResult {
   refreshToken: string;
 }
 
+interface OidcCredentials {
+  silentRefresh: boolean;
+}
+
 export interface OidcAuthFlow {
   refresh: () => Promise<AuthenticatorResult>;
 }
 
 export class OidcAuthenticator {
   private readonly http: HttpClient;
-  private readonly creds: any;
+  private readonly creds: OidcCredentials;
   private accessToken: string;
   private refreshToken?: string;
   private expiresAt: number;
   private refreshRunning: boolean;
+  private refreshInterval!: NodeJS.Timeout;
 
   constructor(http: HttpClient, creds: any) {
     this.http = http;
@@ -57,10 +62,7 @@ export class OidcAuthenticator {
       this.accessToken = resp.accessToken;
       this.expiresAt = resp.expiresAt;
       this.refreshToken = resp.refreshToken;
-      if (!this.refreshRunning && this.refreshTokenProvided()) {
-        this.runBackgroundTokenRefresh(authenticator);
-        this.refreshRunning = true;
-      }
+      this.startTokenRefresh(authenticator);
     });
   };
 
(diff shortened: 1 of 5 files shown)


src/connection/auth.ts
```diff
@@ -270,15 +296,23 @@ class AccessTokenAuthenticator implements OidcAuthFlow {
...
-export class AuthClientCredentials {
+export class AuthClientCredentials implements OidcCredentials {
   private clientSecret: any;
   private scopes?: any[];
+  public readonly silentRefresh: boolean;
 
   constructor(creds: ClientCredentialsInput) {
     this.clientSecret = creds.clientSecret;
     this.scopes = creds.scopes;
+    // Silent token refresh by default
```
dirkkul, line 310: make one function for this?

src/classifications/scheduler.ts
```diff
@@ -95,23 +95,27 @@ export default class ClassificationsScheduler extends CommandBase {
 
   pollForCompletion = (id: any): Promise<Classification> => {
```
dirkkul, line 96: Why does classifcation have its own timeout check?

src/connection/auth.ts
```diff
@@ -270,15 +296,23 @@ class AccessTokenAuthenticator implements OidcAuthFlow {
...
+    if (creds.silentRefresh === undefined) {
```
dirkkul, line 311: just to be sure:
This means that this parameter is not required and this is not a breaking change?

src/classifications/scheduler.ts
parkerduckworth: This has been in the client for a long time. This is the implementation of an option which tells the client to block until classification is complete.

src/connection/auth.ts
parkerduckworth: Done

Issue #17, opened 2023-01-19
Add wait_until_weaviate_is_ready function


dirkkul:
Add a function that checks if weaviate is responding on the given url:port and blocks until it does or a given timeout has passed

dirkkul:
this did not actually help in python - will not do it

PR #34, opened 2023-04-03
Support 'latest' version, explicit binary URL for Embedded DB

parkerduckworth:
- Allow `'latest'` version to automatically download the latest Weaviate server version
- Support explicit URL to weaviate binary in the form of `binaryUrl` property of `EmbeddedOptions`
- Detect machine architecture to download correct binary

### DIFF
--- a/src/embedded/index.ts
+++ b/src/embedded/index.ts
@@ -265,6 +265,7 @@ export class EmbeddedDB {
         })
           .on('finish', () => {
             tarball.close();
+            fs.unlinkSync(tarballPath);
             fs.renameSync(join(dirname(this.options.binaryPath), 'weaviate'), this.options.binaryPath);
             resolve(null);
           })
--- a/src/embedded/index.ts
+++ b/src/embedded/index.ts
@@ -10,7 +10,7 @@ import { extract } from 'tar';
 
 const defaultBinaryPath = join(homedir(), '.cache/weaviate-embedded');
 const defaultPersistenceDataPath = join(homedir(), '.local/share/weaviate');
-const defaultVersion = '1.18.0';
+const defaultVersion = '1.18.1';
 
 interface EmbeddedOptionsConfig {
   host?: string;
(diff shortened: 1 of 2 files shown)


src/embedded/index.ts
```diff
@@ -67,11 +67,15 @@ export class EmbeddedOptions {
...
+      return 'latest';
+    }
     if (cfg.version.match(/[1-9]\.[1-9]{2}\..*/g)) {
       return cfg.version;
-    } else {
-      throw new Error(`invalid version: ${cfg.version}. version must resemble '{major}.{minor}.{patch}'`);
     }
+    throw new Error(
+      `invalid version: ${cfg.version}. version must resemble '{major}.{minor}.{patch}, or 'latest'`
+    );
```
dirkkul, line 78: I think we still should support URLs. In the future we might build binaries for each PR

src/embedded/index.ts
```diff
@@ -145,80 +151,173 @@ export class EmbeddedDB {
...
-        strict: true,
-      })
-        .on('finish', () => {
-          tarball.close();
-          fs.renameSync(join(dirname(opt.binaryPath), 'weaviate'), opt.binaryPath);
-          resolve(null);
+  private buildBinaryUrl(): string {
+    let arch: string;
+    switch (process.arch) {
+      case 'arm64':
```
dirkkul, line 242: did you test this?
On python the platform command returns aarch64 (which is taken from uname), no idea what the TS library does

(Afaik we do not have anyone with an ARM64 laptop)

src/embedded/index.ts
```diff
@@ -145,80 +151,173 @@ export class EmbeddedDB {
...
+    return new Promise((resolve, reject) => {
+      tarball.pipe(
+        extract({
+          cwd: dirname(tarballPath),
+          strict: true,
         })
-        .on('error', function (err) {
-          reject(new Error(`failed to untar binary: ${JSON.stringify(err)}`));
+          .on('finish', () => {
+            tarball.close();
```
dirkkul, line 267: delete tarbal after unpacking? :)

src/embedded/index.ts
```diff
@@ -145,80 +151,173 @@ export class EmbeddedDB {
...
+            resolve(null);
+          })
+          .on('error', (err) => {
+            reject(new Error(`failed to untar binary: ${JSON.stringify(err)}`));
+          })
+      );
+    });
+  }
+
+  private waitTillListening(): Promise<null> {
```
dirkkul, line 278: just a note, this function could be used to implement [this here](https://github.com/weaviate/typescript-client/issues/17)
(obviously should not be part of this PR)

src/embedded/index.ts
parkerduckworth: Yes, I have tested on both arm64 and amd64 environments.

Also see: https://nodejs.org/api/process.html#processarch

src/embedded/index.ts
parkerduckworth: done

PR #35, opened 2023-04-03
Support generative GQL search

parkerduckworth:


src/graphql/getter.test.ts
```diff
@@ -1176,3 +1176,99 @@ describe('hybrid valid searchers', () => {
...
 });
+
+describe('generative search', () => {
+  const mockClient: any = {
+    query: jest.fn(),
+  };
+
+  test('singlePrompt', () => {
+    const expectedQuery =
+      '{Get{Mammal(generate(singleResult:{prompt:"When did dogs become mans best friend?"})' +
```
etiennedi, line 1187: ❤️

etiennedi (approved):
LGTM

src/graphql/generate.ts
```diff
@@ -0,0 +1,49 @@
...
+    this.singlePrompt = args.singlePrompt;
+  }
+
+  toString(): string {
+    this.validate();
+
+    let str = 'generate(';
+    const results = ['error'];
+    if (this.singlePrompt) {
+      str += `singleResult:{prompt:"${this.singlePrompt.replace(/[\n\r]+/g, '')}"}`;
```
antas-marcin, line 27: In other clients we have methods that safely quote text when we build the argument, do we have something similar here?

parkerduckworth: Luckily, like the python client, this client is unaffected by this problem

PR #36, opened 2023-04-07
Export ConsistencyLevel type

antas-marcin:


src/index.ts
```diff
@@ -111,3 +111,4 @@ export * from './backup';
 export * from './cluster';
 export * from './connection';
 export * from './embedded';
+export * from './data/replication';
```
parkerduckworth, line 114: since `./data` is already being exported in this file, can you instead re-export `ConsistencyLevel` from within the `data` module? Like this:

src/data/index.ts:
```ts
// <all the other re-exports>
export type { ConsistencyLevel } from './replication';
```

Then, the `ConsistencyLevel` type will still be available like this:
```ts
import weaviate, {  ConsistencyLevel } from 'weaviate-ts-client';
```

parkerduckworth (changes requested):

parkerduckworth (approved):

PR #56, opened 2023-05-12
gh-43: Full support for ESM modules.

iSuslov:
Fixes https://github.com/weaviate/typescript-client/issues/43 in `v1.3`

**Problem:**
In typescript when using more strict `moduleResolution` such as  `nodenext` or `node16`, typescript can't find type declarations of the esm module which has `.mjs` extension. This is because TypeScript ignores `.js` when there is an identically named `.d.ts` file beside it. Same happens with `.mjs` and `.d.mts` but if no such pair found, typescript thinks there are no type declarations. 
[More info in my comment here.](https://github.com/weaviate/typescript-client/pull/46#issuecomment-1545072740)

**Solution:**
Assuming that weaviate team doesn't want to change package module type from `commonjs` to `module`, keeping `commonjs` as default option, the smallest change is to generate `d.mts` declaration file. Since we don't have actual `.mts` files, it's not easy to emit one using standard bundlers and builders, [see](https://github.com/microsoft/TypeScript/issues/49462). The best workaround I see is to clone original type definitions after build. This will make this PR small and if one day weaviate team decides to implement a build pipeline it will be straightforward to implement this step using other tools.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,9 +7,9 @@
   "types": "./dist/index.d.ts",
   "exports": {
     ".": {
+      "types": "./dist/index.d.ts",
       "require": "./dist/index.js",
-      "import": "./dist/index.mjs",
-      "types": "./dist/index.d.ts"
+      "import": "./dist/index.mjs"
     }
   },
   "engines": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -7,9 +7,9 @@
   "types": "./dist/index.d.ts",
   "exports": {
     ".": {
+      "types": "./dist/index.d.ts",
```
iSuslov, line 10: `types` should always go first. [See typescript doc]( https://www.typescriptlang.org/docs/handbook/esm-node.html#packagejson-exports-imports-and-self-referencing )

package.json
```diff
@@ -18,11 +18,12 @@
...
     "/dist/index.js",
     "/dist/index.mjs",
-    "/dist/index.d.ts"
+    "/dist/index.d.ts",
+    "/dist/index.d.mts"
   ],
   "scripts": {
     "test": "tsc -noEmit -p tsconfig-test.json && jest --useStderr --runInBand --detectOpenHandles",
-    "build": "npm run lint && tsup src/index.ts --format cjs,esm --dts --clean --platform neutral --minify",
+    "build": "npm run lint && tsup src/index.ts --format cjs,esm --dts --clean --platform neutral --minify && cp ./dist/index.d.ts ./dist/index.d.mts",
```
iSuslov, line 26: Simply duplicate declaration file with `d.mts` extension. This will make typescript compiler happy because for `.mjs` module types will be visible.

parkerduckworth:
@iSuslov the intention of making this package a CommonJS library is to keep backwards compatibility with the projects who previously used the CommonJS JavaScript client which this client was migrated from. 

Does changing this package type to module break this compatibility?

tsmith023:
@iSuslov, what is the status of this PR? Are you in a position to reply to https://github.com/weaviate/typescript-client/pull/56#issuecomment-1545785442?

PR #58, opened 2023-05-22
UUIDv5 support

parkerduckworth:
Introduces a `generateUuid5` function, which takes in any identifier and an optional namespace, to generate deterministic UUIDs.

Example: 
```ts
import { generateUuid5 } from 'weaviate-ts-client';

const withNamespace = generateUuid5('some-string-identifer', 'some-namespace');
console.log(withNamespace);

const withoutNamespace = generateUuid5('some-string-identifer');
console.log(withoutNamespace);
```

### DIFF
--- a/src/utils/uuid.ts
+++ b/src/utils/uuid.ts
@@ -1,0 +1,8 @@
+import { v5 as uuid5 } from 'uuid';
+
+// Generates UUIDv5, used to consistently generate the same UUID for
+// a specific identifier and namespace
+export function generateUuid5(identifier: any, namespace: any = ''): string {
+  const stringified = JSON.stringify(identifier) + JSON.stringify(namespace);
+  return uuid5(stringified, uuid5.DNS).toString();
+}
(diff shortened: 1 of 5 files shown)


src/utils/uuid.ts
```diff
@@ -0,0 +1,8 @@
+import { v5 as uuid5 } from 'uuid';
+
+// Generates UUIDv5, used to consistently generate the same UUID for
+// a specific identifier and namespace
+export function generateUuid5(identifier: any, namespace: any = ''): string {
```
antas-marcin, line 5: just a suggestion, maybe just `uuid5` would be sufficient? I'm OK with both :)

antas-marcin: and args can be `any`? or only `strings`?

parkerduckworth: I have updated the param type for both `identifier` and `namespace` to `string | number`

antas-marcin (approved):

dandv:
Are there plans to support object parameters, for feature parity with the [Python equivalent](https://weaviate-python-client.readthedocs.io/en/stable/weaviate.util.html#weaviate.util.generate_uuid5)?

PR #67, opened 2023-06-08
Add support for multi tenancy endpoints

antas-marcin:


src/schema/tenantsGetter.ts
```diff
@@ -0,0 +1,36 @@
...
+import { Tenant } from '../openapi/types';
+
+export default class TenantsGetter extends CommandBase {
+  private className?: string;
+
+  constructor(client: Connection) {
+    super(client);
+  }
+
+  withClassName = (className: string) => {
```
dirkkul, line 13: if this is required, doesn't it make more sense to have it directly in the constructor? (If that is possible)

src/schema/tenantsCreator.ts
```diff
@@ -0,0 +1,49 @@
...
+    if (!isValidStringProperty(this.className)) {
+      this.addError('className must be set - set with .withClassName(className)');
+    }
+  };
+
+  validateTenants = () => {
+    if (!this.tenants || this.tenants.length == 0) {
+      this.addError('tenants must be set - set with .withTenants(tenants)');
+    }
+  };
```
dirkkul, line 34: same here, if it is required, let's just add it to the constructor

src/schema/tenantsDeleter.ts
```diff
@@ -0,0 +1,49 @@
...
+    if (!isValidStringProperty(this.className)) {
+      this.addError('className must be set - set with .withClassName(className)');
+    }
+  };
+
+  validateTenants = () => {
+    if (!this.tenants || this.tenants.length == 0) {
+      this.addError('tenants must be set - set with .withTenants(tenants)');
+    }
+  };
```
dirkkul, line 34: same

dirkkul (approved):

aliszka (approved):

src/schema/tenantsGetter.ts
antas-marcin: fixed

Issue #16, opened 2023-02-20
Add method to check if a class exists


dirkkul:
Add a function that returns if a given class exists in the schema

Proposal signature (copied from python as I don't know JS):
`client.schema.exists(className: str) ->bool`

tsmith023:
@dirkkul, introduced by https://github.com/weaviate/typescript-client/pull/70 :grin:

PR #70, opened 2023-06-15
Add method to check if class exists

tsmith023:
This PR introduces a method to the `schema` property of `client` that checks for the existence of a class within the database by its `className`.

In an effort to keep the implementation decoupled but logical, the `ClassExists` builder class is introduced that conforms to the logic of other builder classes within the `src/schema/` directory. The `do()` method of this class is, however, abstracted away from the end user such that they only access:
```typescript
const exists: boolean = await client
  .schema
  .exists('MyClass')
```
without the added overhead and complexity of making many builder calls to achieve the result. This was done to conform with the previously raised issue: https://github.com/weaviate/typescript-client/issues/16.

The implementation simply calls the `/schema` endpoint of the Weaviate REST API and uses the `.some()` `Array.prototype` method to compare the returned `classes` property with the provided `className`.

@dirkkul :grin:

### DIFF
--- a/src/schema/classExists.ts
+++ b/src/schema/classExists.ts
@@ -1,0 +1,38 @@
+import { isValidStringProperty } from '../validation/string';
+import Connection from '../connection';
+import { CommandBase } from '../validation/commandBase';
+import { WeaviateSchema } from '../openapi/types';
+
+export default class ClassExists extends CommandBase {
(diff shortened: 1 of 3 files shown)


parkerduckworth (approved):
Very nice, thanks Tommy! 🙂

Issue #63, opened 2023-06-02
client.data.checker() returns a function instead of a boolean


dandv:
```js
const client = weaviate.client({
  scheme: 'https',
  host: 'edu-demo.weaviate.network',
  apiKey: new weaviate.ApiKey('learn-weaviate'),
});

const id = '00ff6900-e64f-5d94-90db-c8cfa3fc851b';

result = await client.data
  .checker()
  .withClassName('JeopardyQuestion')
  .withId(id)
  .do();

console.log(typeof result, `result: ${result}`, JSON.stringify(result));
```

The output will be this,

> function result: t=>{if(t.status>=400)return t.text().then(e=>{let r;try{r=JSON.stringify(JSON.parse(e))}catch{r=e}return Promise.reject(new Error(`usage error (${t.status}): ${r}`))});if(i)return t.json()} undefined

The object does exist,

```ts
result = await client.data
  .getterById()
  .withClassName('JeopardyQuestion')
  .withId(id)
  .do();

console.log(JSON.stringify(result, null, 2));
```

BTW the [tests for `checker()`](https://github.com/weaviate/typescript-client/blob/main/src/data/journey.test.ts) lack `.withClassName()`, which results in a warning, "Usage of objects paths without className is deprecated in Weaviate 1.19.6. Please provide className parameter".

tsmith023:
@dandv, this seems to be a multi-faceted issue. The core problem you're seeing is that the `learn-weaviate` API key in fact lacks the permissions necessary to use the `.checker()` method on the `https://edu-demo.weaviate.network` URL. This is because `.checker()` makes an HTTP HEAD request under-the-hood and this apparently requires elevated access.

You were unable to see this error in the client, however, due to the issue that you raised with regards to the return type in your MRE. A solution to this is open in a PR: https://github.com/weaviate/typescript-client/pull/71, but the fundamental problem will remain: the demo API key cannot make HEAD requests so cannot use the `.checker()` method.

[...]

PR #71, opened 2023-06-17
Hotfix client data checker return

tsmith023:
This PR fixes a bug raised in https://github.com/weaviate/typescript-client/issues/63 where a >400 error code returned from a HEAD request does not propagate correctly to the user. Also contained are small type improvements to help with DevEx and increased ease of understanding of the codebase.

@parkerduckworth :grin:


parkerduckworth:
Thanks @tsmith023 !

PR #74, opened 2023-07-06
Add hybrid fusion type support

antas-marcin:


src/graphql/journey.test.ts
```diff
@@ -372,6 +372,36 @@ describe('the graphql journey', () => {
       });
   });
 
+  test('graphql get hybrid with query, alpha, properties and fushionType: rankedFusion', () => {
+    return client.graphql
+      .get()
+      .withClassName('Article')
+      .withHybrid({ query: 'Apple', properties: ['title'], alpha: 0, fusionType: 'rankedFusion' })
```
dirkkul, line 379: fusionType is an grapql enum and not a string, so the '' are not needed

antas-marcin: done

dirkkul (approved):

PR #76, opened 2023-07-06
Add class parameter to nodes API

antas-marcin:


src/cluster/journey.test.ts
```diff
@@ -20,15 +20,16 @@ describe('cluster nodes endpoint', () => {
     return client.cluster
       .nodesStatusGetter()
       .do()
-      .then((nodesStatusResponse: any) => {
+      .then((nodesStatusResponse: NodesStatusResponse) => {
         expect(nodesStatusResponse.nodes).toHaveLength(1);
-        const node = nodesStatusResponse.nodes[0];
+        const node = nodesStatusResponse.nodes![0];
```
dirkkul, line 25: what do these warnings mean?

>Forbidden non-null assertion

dirkkul (approved):

src/cluster/journey.test.ts
antas-marcin: `nodesStatusResponse.nodes` is optional (as we now use a type in response) and that's why we get those warnings, but let me write this code in a "safe" way

aliszka (approved):

PR #81, opened 2023-08-05
instance_version_warning

Abhijithchintu:


Abhijithchintu:
I agree to CLA

src/validation/version.ts
```diff
@@ -0,0 +1,11 @@
...
+    if (typeof version === 'string') {
+        const versionNumbers = version.split('.');
+        if (versionNumbers.length >= 2) {
+          const major = parseInt(versionNumbers[0], 10);
+          const minor = parseInt(versionNumbers[1], 10);
+          return (major <= 1 && minor < 16);
+        }
+    }
+    return false;
+}
```
tsmith023, line 11: The boolean logic here is a little confusing since the function is named `isValidWeaviateVersion` yet it returns a falsey value on successful completion and truthy value on error/issue. With this logic, I would expect the function to be called `isNotValidWeaviateVersion`.

I think it would be better if this name was maintained but the internal logic was flipped so that an invalid version returns a falsey value.

src/utils/beaconPath.ts
```diff
@@ -41,6 +42,11 @@ export class BeaconPath {
     } else {
       support.warns.notSupportedClassNamespacedEndpointsForBeacons();
     }
+    if(support.version){
+      if (isValidWeaviateVersion(support.version)){
```
tsmith023, line 46: If `isValidWeaviateVersion` changes as per my comment below, this will need to change too 😁

src/data/path.ts
```diff
@@ -176,6 +177,11 @@ export class ReferencesPath {
       } else {
         support.warns.notSupportedClassNamespacedEndpointsForReferences();
       }
+      if(support.version){
+        if (isValidWeaviateVersion(support.version)){
```
tsmith023, line 181: If `isValidWeaviateVersion` changes as per my comment below, this will need to change too 😁

tsmith023 (changes requested):
Thanks for the contribution and for developing it inline with the established style of the codebase! There's only one major comment on the `isValidWeaviateVersion` function otherwise it looks good to me 😁

Abhijithchintu:
Thanks for reviewing and suggesting changes, i have written function and then changed logic again. Hence it contradicted with function name, later i forgot to change function name. Now i have made the changes along with lint fix. Please review the pr again.

src/validation/version.ts
```diff
@@ -0,0 +1,11 @@
...
+  if (typeof version === 'string') {
+    const versionNumbers = version.split('.');
+    if (versionNumbers.length >= 2) {
+      const major = parseInt(versionNumbers[0], 10);
+      const minor = parseInt(versionNumbers[1], 10);
+      return major <= 1 && minor < 16;
+    }
+  }
+  return false;
+}
```
tsmith023, line 11: Sorry for not being clearer in the original review but I still think this is confusing since a double negative is involved in the logic & naming convention here. Here we are saying "Is the Weaviate version invalid" and return "No" on a success, which is easy to get confused about due to the double negative

I think naming it `isValidWeaviateVersion`, returning `true` on line 10, and refactoring line 7 to return `false` is preferable. Perhaps the following on line 7:
```typescript
return major <= 1 && minor < 16 ? false : true;
```

Then the conditional statements in `path.ts` and `beaconPath.ts` would be:
```typescript
if(!isValidWeaviateVersion(support.version)) { ... }
```

In this way double negatives are avoided and the code is easier to read 😁 The logic of saying "Is not valid" in the conditional statements is still clear from the `!` operator also

tsmith023 (changes requested):

tsmith023:
@Abhijithchintu, interestingly `eslint` isn't a fan of my recommendation 😅 I think this:
```typescript
return major <= 1 && minor < 16 ? false : true;
```
needs changing to this:
```typescript
return !(major <= 1 && minor < 16)
```
Make sure to the run the linter before you push the changes to ensure my suggestion is accepted by it!

PR #83, opened 2023-08-17
Activate deactivate tenants support

aliszka:


src/schema/tenantsUpdater.ts
```diff
@@ -0,0 +1,22 @@
+import Connection from '../connection';
+import { CommandBase } from '../validation/commandBase';
+import { Tenant } from '../openapi/types';
+
+export default class TenantsCreator extends CommandBase {
```
tsmith023, line 5: Should this be `TenantsUpdater`?

package.json
```diff
@@ -68,6 +69,7 @@
     "husky": "^8.0.3",
     "jest": "^29.4.3",
     "lint-staged": "^13.2.0",
+    "openapi-typescript": "^5.4.1",
```
antas-marcin, line 72: I would not make this `dev dependency` upgradable, I would just set it to `5.4.1` bc higher versions don't support Swagger 2.0

tools/refresh_schema.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+set -euo pipefail
+
+var="${1:-master}"
```
antas-marcin, line 5: I would change the `var` name to `branch`

tsmith023 (approved):
LGTM :)

src/schema/tenantsUpdater.ts
aliszka: Right, copy/paste 🤦

package.json
aliszka: should be fine. `^` allows just for minor and patch updates.

antas-marcin (approved):

Issue #87, opened 2023-08-20
Connection fails when user-provided host URL contains the scheme


thomashacker:
### Issue:
When initializing the Weaviate Client, if a user provides a host URL that already contains a scheme (e.g., https), the connection fails. This is because both the host and scheme are concatenated in the code.

### Expected Behavior:
The client should handle cases where the scheme might already exist in the provided host URL. The goal is consistent behavior in both Python and Typescript clients. For instance, a user should be able to use the same URL for both the Python and TypeScript clients without adjusting the URL format for each.

### How to Reproduce:

```
# Python Example

client = weaviate.Client(url = "https://some-endpoint.weaviate.network")
```

```
// Works

const client: WeaviateClient = weaviate.client({
  scheme: 'https', host: 'some-endpoint.weaviate.network'
});
```

```
// Fails

const client: WeaviateClient = weaviate.client({
  scheme: 'https', host: 'https://some-endpoint.weaviate.network'
});
```

PR #88, opened 2023-08-20
Handle existing scheme in host

thomashacker:
This `PR` aims to fix https://github.com/weaviate/typescript-client/issues/87 and adds a simple check before creating the `baseUrl`

```
const version = '/v1';
const baseUri = config.host.startsWith(`${config.scheme}://`)
    ? `${config.host}${version}`
    : `${config.scheme}://${config.host}${version}`;
```

This allows users to use a host URL with a scheme if it matches the specified scheme.

The PR also adds two unit tests to verify that it constructs the correct URLs.

### DIFF
--- a/src/connection/unit.test.ts
+++ b/src/connection/unit.test.ts
@@ -124,6 +124,40 @@ describe('mock server auth tests', () => {
     await conn.login().then((key) => expect(key).toEqual(apiKey));
   });
 
+  it('should construct the correct baseUri for conn.http when host contains scheme', async () => {
+    const apiKey = 'abcd123';
+
+    const conn = new Connection({
+      scheme: 'http',
+      host: 'http://localhost:' + server.port,
+      apiKey: new ApiKey(apiKey),
+    });
+
+    await conn.http.get('/testEndpoint');
+    const lastRequestURL = server.lastRequest().path; // Get the path of the last request
+    const expectedPath = '/v1/testEndpoint';
+    expect(lastRequestURL).toEqual(expectedPath);
+  });
+
+  it('should construct the correct baseUri for conn.query when host contains scheme', async () => {
+    const apiKey = 'abcd123';
+
+    const conn = new Connection({
+      scheme: 'http',
+      host: 'http://localhost:' + server.port,
+      apiKey: new ApiKey(apiKey),
+    });
+
+    try {
+      await conn.query('{ query { users } }');
+    } catch (error) {
+      // We just want to test the last request path
+    }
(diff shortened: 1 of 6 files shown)


tsmith023 (commented):
Thanks for the PR @thomashacker 😁

Shall we make the `config.host` parameter optional now that users are free to supply it within the host? Then the scheme would be pulled out of `config.host` by splitting the string on `://`. If they don't provide `config.scheme` and it cannot be found in `config.host`, then an error is thrown.

I'm wondering if it would ever be confusing for users coming from Python to specify the URL in `config.host` including the protocol only to then have to specify it in `config.scheme` also. What do you think?

thomashacker:
I agree,

As far as I can see, the `scheme` parameter is only being used to build the `baseURL` for both the `HTTPClient` and `GQLClient`. As you suggested, If the scheme is already present in the host, we can make specifying the scheme optional.

I would also move the whole `scheme`/`host` processing upwards, I'll look into it!

thomashacker:
- Made the `scheme` parameter optional and removed `throw error` when it doesn't exist
- Moved `schema` and `host` checks to `Connection`, adding following unit tests:
1. Host contains scheme and it's the same as specified (Use host only)
2. Host contains scheme but it's not the same as specified (throw error)
3. Host contains scheme and scheme was not specified (Use host only)
4. Host does not contain scheme and no scheme was specified (throw error)
5. Host does not contain scheme but a scheme was specified (default)

- Both the `HttpClient` and `GQLClient` are now using `params.host` parameter + adding their prefixes

src/connection/unit.test.ts
```diff
@@ -124,6 +124,75 @@ describe('mock server auth tests', () => {
...
+    const createConnection = () => {
+      return new Connection({
+        scheme: 'https',
+        host: 'http://localhost:' + server.port,
+        apiKey: new ApiKey(apiKey),
+      });
+    };
+
+    expect(createConnection).toThrow(
+      'The host contains a different protocol than specified in the scheme (scheme: https:// != host: http://)'
```
tsmith023, line 177: On a brief glance, I got a bit confused by this error message. Might it be more accurate to say:
`'The host contains a different protocol than specified in the scheme (scheme: https != host: http)'`?

I can foresee a user seeing this error and trying to add `https://` to the `scheme` field to fix it

tsmith023 (approved):
LGTM otherwise :)

src/connection/unit.test.ts
thomashacker: Good idea, I changed the error message to make it more clear!

PR #89, opened 2023-08-21
Add nearMedia filters for multi2vec-bind model

tsmith023:
This PR introduces functionality to filter based on media type when using the multi2vec-bind vectorisation model. In doing so, it attempts to refactor the `nearImage` filter implementation while leaving the current API unchanged to preserve BC.


src/graphql/getter.ts
```diff
@@ -133,21 +144,68 @@ export default class GraphQLGetter extends CommandBase {
...
     }
-
     return this;
   };
 
+  withNearImage = (args: NearImageArgs) => {
+    return this.withNearMedia({
+      ...args,
+      type: NearMediaType.Image,
+      media: args.image ? args.image : 'UNSET',
```
antas-marcin, line 165: why only for `NearImage` with have this check of `image` existance then if not set we set `UNSET` value?

tsmith023: It's because of the need for backwards compatability with the `NearImageArgs` interface. Since it is defined with `image?: string` users can provide `null` or `undefined`. But ideally, it should always be a required field since it is fundamental to the `.with_` method.

So, if it's not provided it gets assigned to `'UNSET'` to be used when performing validation on the `toString()` method. This convo shows that it's convoluted so I will refactor to it to make it simpler 😁

antas-marcin (approved):

antas-marcin (approved):

src/graphql/getter.test.ts
```diff
@@ -1091,6 +1091,93 @@ describe('nearImage searchers', () => {
...
+      query: jest.fn(),
+    };
+
+    const subQuery = `(nearDepth:{depth:"iVBORw0KGgoAAAANS"})`;
+    const expectedQuery = `{Get{Person` + subQuery + `{name}}}`;
+
+    new Getter(mockClient)
+      .withClassName('Person')
+      .withFields('name')
+      .withNearDepth({ depth: 'iVBORw0KGgoAAAANS' })
```
antas-marcin, line 1157: you can also a test (I would say that 1 would be enough) that shows that `distance` and `certainty` also is being properly serialized

tsmith023: Will do!

PR #96, opened 2023-09-26
fix: pass-through tenant when batch-creating cross-references

flipace:
When using multi tenancy and trying to use the cross-reference batcher, the tenant would not be passed through.
This caused an error because weaviate expects the tenant for tenant-enabled classes.


tsmith023 (approved):
Thanks for the PR @flipace! Since it's such a small change, there's no need to add tests and I'll merge right away 😁

PR #102, opened 2023-11-24
Collections v3 client

tsmith023:
This PR introduces the `v3` collections client as a complete rewrite of the API surface to match the `PYv4` client's new collections API. It makes use of all three (`REST`, `GQL`, `gRPC`) of Weaviate's APIs when offering its functionality. As such, there are several instances where this API depends on the builders of the previous API, especially when working with functionality that is only available through the `REST` or `GQL` APIs.

Key features added in this API are:
- The `.collections` namespace within the `client` object from which all the new functionality is accessible, e.g. `client.collections.get('MyCollection')`
- The `Collection` object within which exist several namespaces for accessing key core Weaviate functionality:
    - `.aggregate` for performing queries using the `Aggregate` `/v1/graphql` query
    - `.backup` for managing backups through the `/v1/backups` endpoint
    - `.config` for managing a collection's configuration through the `/v1/schema` endpoint
    - `.data` for managing a collection's data objects, e.g. CUD ops, through the `/v1/objects` endpoint, the `weaviate.v1.weaviate.BatchObjects` gRPC method (for `.insertMany`), and the `weaviate.v1.weaviate.BatchDelete` gRPC method (for `.deleteMany`)
    - `.filter` for defining fully generic filters for queries and `.data.deleteMany` that provides type safety given a user-provided generic
    - `.generate` for querying objects from a collection with generative capabilities through the `weaviate.v1.weaviate.Search` gRPC method
    - `.metrics` for defining fully generic metrics objects for aggregations that provides type safety given a user-provided generic
    - `.query` for querying objects from a collection without generative capabilities through the `weaviate.v1.weaviate.Search` gRPC method
    - `.sort` for defining fully generic sorting constraints for queries that provides type safety given a user-provided generic
    - `.tenants` for managing the tenants of a collection through `/v1/tenants` endpoint and the `weaviate.v1.weaviate.TenantsGet` gRPC method

The new API is also fully generic meaning that there are no instances of `Record<string, Any>` in the codebase (and if there are this is a mistake!). Users should have fine-grained control over the types they use when working with the API and will be able to define their own complex typing systems on top of the client.


src/collections/aggregate/integration.test.ts
```diff
@@ -0,0 +1,382 @@
...
+    //     collection.metrics.aggregate('number').number(),
+    //     collection.metrics.aggregate('numbers').number(),
+    //     collection.metrics.aggregate('date').date(),
+    //     collection.metrics.aggregate('dates').date(),
+    //     collection.metrics.aggregate('boolean').boolean(),
+    //     collection.metrics.aggregate('booleans').boolean(),
+    //     // Metrics.aggregate('ref').reference(['pointingTo'])
+    //   ],
+    // });
+    // expect(resDef).toEqual(result); // leads to flakes due to weaviate's behaviour when calculating modes of dates
```
parkerduckworth, line 224: Is this still needed?

parkerduckworth (approved):

src/collections/aggregate/integration.test.ts
tsmith023: Good spot, will remove! I left to identify the flake but it's very low prio

Issue #113, opened 2024-03-06
Hybrid search is missing target vectors


joshdschneider:
I'm unable to pass target vectors in a hybrid search query. Here's an example:

```
const result = await client.graphql
  .get()
  .withClassName('WineReviewNV')
  .withHybrid({
    query: 'a sweet German white wine',
    targetVectors: ['title_country'],
  })
  .withLimit(2)
  .withFields('title review_body country')
  .do();

console.log(JSON.stringify(result, null, 2));
```

It looks like it's missing from the `GraphQLHybrid` class. I'll go ahead and open a PR to fix.

PR #114, opened 2024-03-06
Add support for target vectors in hybrid queries

joshdschneider:
Fixes [113](https://github.com/weaviate/typescript-client/issues/113).

### DIFF
--- a/src/graphql/hybrid.ts
+++ b/src/graphql/hybrid.ts
@@ -3,6 +3,7 @@ export interface HybridArgs {
   query: string;
   vector?: number[];
   properties?: string[];
+  targetVectors?: string[];
   fusionType?: FusionType;
 }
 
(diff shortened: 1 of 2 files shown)


tsmith023:
Hi @joshdschneider, thanks for the PR on this one! Could you please add an integration test within the describe block [here](https://github.com/weaviate/typescript-client/blob/main/src/graphql/journey.test.ts#L2106) for your changes, cheers! 😁

tsmith023:
@joshdschneider, your test fails (in quite an interesting way). It would seem that hybrid, due to its mixed nature as bm25 + vector search with the fusion algo, performs some cutoff returning only one result for you query!

joshdschneider:
@tsmith023 ah, I see. Sorry about that—fixed!

tsmith023 (approved):
Thanks again!

PR #122, opened 2024-04-08
Respond to beta user feedback

tsmith023:
- Convert `collections.get()` to return a Promise that checks for existence


tsmith023:
This use-case will be met by the `collection.exists()` method. `.get()` will be renamed to `.use()` to avoid confusion around its nomenclature in relation to HTTP GET

PR #160, opened 2024-07-04
Add support for the kagome_kr tokenizer

databyjp:
Adds support for the `kagome_kr` tokenizer (to be added in `1.26`)

https://github.com/weaviate/weaviate/pull/5211/files


tsmith023 (approved):
Thanks for the PR 😁

PR #196, opened 2024-09-18
💅 Update `text2vec-azure-openai` to utilize `isAzure: true` flag and mark `resourceName` + `deploymentId` as optional

flipace:
> This relates to the changes in https://github.com/weaviate/weaviate/pull/5776

With this adjustment, devs can use the `text2VecAzureOpenAI` vectorizer, without specifying `deploymentId` or `resourceName` upfront for their collection.

Instead, they may provide the headers `X-Azure-Deployment-Id` and `X-Azure-Resource-Name` in their requests to set these.

```ts
vectorizer.text2VecAzureOpenAI({
  vectorizeCollectionName: false,
})
```

Internally, using text2VecAzureOpenAI will set the an `isAzure: true` flag for the OpenAI vectorizer, so it understands that the Azure logic must be used.


src/collections/configure/unit.test.ts
```diff
@@ -597,6 +597,24 @@ describe('Unit testing of the vectorizer factory class', () => {
...
+    expect(config).toEqual<VectorConfigCreate<never, undefined, 'hnsw', 'text2vec-azure-openai'>>({
+      name: undefined,
+      vectorIndex: {
+        name: 'hnsw',
+        config: undefined,
+      },
+      vectorizer: {
+        name: 'text2vec-azure-openai',
+        config: {
+          isAzure: true,
```
tsmith023, line 617: Is the output the same if a user supplies `configure.vectorizer.text2VecAzureOpenAI()` without an empty object argument?

src/collections/config/types/vectorizer.ts
```diff
@@ -180,10 +180,12 @@ export type Text2VecAWSConfig = {
...
-  /** The deployment ID to use */
-  deploymentId: string;
-  /** The resource name to use. */
-  resourceName: string;
+  /** The deployment ID to use. If left empty, must be provided via X-Azure-Deployment-Id header  */
+  deploymentId?: string;
+  /** The resource name to use. If left empty, must be provided via X-Azure-Resource-Name header */
+  resourceName?: string;
+  /** Will automatically be set to true. You don't need to set this manually. */
+  isAzure?: true;
```
tsmith023, line 188: If this is always `true`, does it still need the `?` operator? If not, can we remove it?

src/collections/config/types/generative.ts
```diff
@@ -75,8 +75,9 @@ export type GenerativeOpenAIConfig = GenerativeOpenAIConfigBase & {
 };
 
 export type GenerativeAzureOpenAIConfig = GenerativeOpenAIConfigBase & {
-  resourceName: string;
-  deploymentId: string;
+  resourceName?: string;
+  deploymentId?: string;
+  isAzure?: true;
```
tsmith023, line 80: If this is always true, does it still need the ? operator? If not, can we remove it?

tsmith023 (commented):
Thanks for the contribution! Left a few comments mainly around house-keeping otherwise the PR looks great 😁

src/collections/configure/unit.test.ts
flipace: Yes, it should be - I adjusted the types to allow no config object to be passed 👍

src/collections/config/types/vectorizer.ts
flipace: So the `true` is always set by the text2VecAzureOpenAI function internally, should not be necessary to be passed by the user - but due to how the types in general are structured I could not easily 1) remove it completely from the config object, nor 2) remove the optional operator, since then the user would be required to supply `isAzure: true` manually 🤔

src/collections/config/types/generative.ts
flipace: same as https://github.com/weaviate/typescript-client/pull/196#discussion_r1771153555 🤔

src/collections/config/types/vectorizer.ts
tsmith023: Ahhh, okay I see now. I think this makes sense if the user makes use of the `.azureOpenAI` method but part of the API is still to allow users to work with the raw types if they so wish. As such, if a user did:
```typescript
generative: {
  name: 'generative-openai',
  config: {
    deploymentId: config.deploymentId,
    resourceName: config.resourceName,
     baseURL: config.baseURL,
  }
}
```
then the type system would allow it since `isAzure` is optional yet the runtime would interpret this as `isAzure: undefined`, which is a false-y value.

I like the idea of introducing the `isAzure` flag to the TS client but I think it may be better placed as a pure internal, e.g. not exposed in the user types, wdyt?

src/collections/config/types/generative.ts
tsmith023: Replied below 😁

src/collections/config/types/vectorizer.ts
flipace: I would love to not expose it - but I could not figure out how to make it work so that it is not marked as unexpected internally where i place it 😬 

This part here is what I mean:

CleanShot 2024-10-07 at 11 02 14@2x

I wouldn't know how to tell TS here that this is an expected prop without adding a `// @ts-expect-error` 🤔

tsmith023: My thinking is that `isAzure` removed from the `...Config` types and instead interpreted by the client's runtime itself depending on the name of the module. So this would most likely require the addition of `generative-azure-openai`, alongside `text2vec-azure-openai`, that is then parsed appropriately in the collection creation logic

There we'd have some boolean clauses to determine whether the module is an azure one, based on the name, and then inject `isAzure: true` into the config appropriately. IMO, this be the most consistent for the client/server relationship as I'm sure there will be future refactoring of the server that changes this behaviour. Then, we'd only break the internal relationship rather than the public API

We already do something similar [here](https://github.com/weaviate/typescript-client/blob/main/src/collections/index.ts#L180). wdyt about extending this logic as described above?

If you'd rather not then that's fine, I can add it to my backlog 😁 Also, sorry for the spaghetti of the `collection.create` method, I've not had the chance to refactor it into a better structure 😅

flipace: Got it - this definitely makes more sense 👍 I didn't look into this part.

I'm drowning a bit in other work right now, but I should be able to look into this in more detail hopefully next week or so 🙏

tsmith023: If I get round to it this week, I'll ping you on here to let you know. Thanks for your help so far!

PR #214, opened 2024-10-22
Support new HNSW filtering strategy config options

tsmith023:


src/graphql/journey.test.ts
```diff
@@ -1659,7 +1659,7 @@ describe('query cluster with consistency level', () => {
   });
 });
 
-describe('query with group by', () => {
+describe.skip('query with group by SKIPPED BECAUSE OF XREFS RETURN OPTIMISATION BUG', () => {
```
parkerduckworth, line 1662: Is there a GH issue link we can add here?

src/schema/journey.test.ts
```diff
@@ -816,8 +767,8 @@ async function newClassObject(
       },
       stopwords: {
         preset: 'en',
-        additions: null,
-        removals: null,
+        additions: null as unknown as undefined, // hack to deal with weird typing
+        removals: null as unknown as undefined, // hack to deal with weird typing
```
parkerduckworth, line 771: 😭

parkerduckworth (approved):

src/schema/journey.test.ts
tsmith023: Hahah yeh, it's because the `WeaviateClass` type that comes from the OpenAPI schema does not contain `null` but I want to be able to type this method for ease of use. I will look again at why this is `null` rather than `string[] | undefined` as it is in the `WeaviateClass` type to be sure

src/graphql/journey.test.ts
tsmith023: Not right now, I communicated the problem with Dirk. I will not merge the `dev/1.27` branch until it is fixed in core or we have an issue tracking it. Just added this skip for this PR into `dev/1.27`

src/schema/journey.test.ts
tsmith023: Okay there's no parsing in the TS client; Weaviate returns a `null` here rather than `undefined` despite the typing. How confusing!

PR #236, opened 2024-11-27
Text2vec weaviate

augustas1:
adds `text2vec-weaviate` vectorizer

### DIFF
--- a/src/collections/config/types/vectorizer.ts
+++ b/src/collections/config/types/vectorizer.ts
@@ -40,6 +40,7 @@ export type Vectorizer =
   | 'text2vec-google'
   | 'text2vec-transformers'
   | 'text2vec-voyageai'
+  | 'text2vec-weaviate'
   | 'none';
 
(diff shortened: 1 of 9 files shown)


src/collections/config/types/vectorizer.ts
```diff
@@ -406,6 +407,20 @@ export type Text2VecVoyageAIConfig = {
   vectorizeCollectionName?: boolean;
 };
 
+/**
+ * The configuration for text vectorization using Weaviate's self-hosted text-based embedding models.
+ *
+ * TODO: add documentation reference once available.
+ */
+export interface Text2VecWeaviateConfig {
```
tsmith023, line 415: Can we change this declaration to be `type` rather than `interface` to maintain the convention of the codebase?

tsmith023 (changes requested):
Just a small comment 😁

tsmith023 (approved):

PR #242, opened 2025-01-03
Fix inputs and mappings for JinaAI and VoyageAI modules

tsmith023:


malgamves (approved):
Looks like it's all covered! 🙌

src/collections/configure/vectorizer.ts
```diff
@@ -318,12 +318,15 @@ export const vectorizer = {
       vectorIndexConfig,
       vectorizerConfig: {
         name: 'multi2vec-voyageai',
-        config: {
-          ...config,
-          imageFields: imageFields?.map((f) => f.name),
-          textFields: textFields?.map((f) => f.name),
-          weights: Object.keys(weights).length === 0 ? undefined : weights,
-        },
+        config:
```
malgamves, line 321: Doesn't the API need to know what is an image field and what isn't

PR #246, opened 2025-01-06
Make improvments to hybrid ux

tsmith023:
- add `aggregate.hybrid` method
- support `maxVectorDistance` in this new method
- fix minor typing issues when doing grouped aggregates
- add integration tests for aggregate search queries

Closes: https://github.com/weaviate/typescript-client/issues/219


src/collections/aggregate/index.ts
```diff
@@ -7,26 +7,25 @@ import { FilterValue } from '../filters/index.js';
...
-  groupBy: (keyof T & string) | GroupByAggregate<T>;
+export type AggregateGroupByOptions<T, M> = AggregateBaseOptions<T, M> & {
+  groupBy: T extends undefined ? string : (keyof T & string) | GroupByAggregate<T>;
 };
 
 export type GroupByAggregate<T> = {
-  property: keyof T & string;
+  property: T extends undefined ? string : keyof T & string;
   limit?: number;
 };
```
bevzzz, line 26: A few ideas here:

1. Define `PropertyOf` type to reuse across the lib. The signature of the composed types becomes quite self-documenting (e.g. `AggregateGroupByOptions<T> = { groupBy: PropertyOf<T> }`). Also gives a chance to write a more detailed TSDoc for `PropertyOf`, in case there is something extra the user should know.  

```ts
type PropertyOf<T> = T extends undefined ? string : (keyof T & string)
```

2. Do we need to nest `GroupByAggregate`? Can this work:

```ts
type AggregateGroupByOptions<T, M> = AggregateBaseOptions<T, M> & {
  groupBy: PropertyOf<T>;
  limit?: number;
};
```

or does `groupBy.limit` has a different meaning such that it only makes sense nested? I browsed through [the docs](https://weaviate.io/developers/weaviate/api/graphql/aggregate) but couldn't find a good example of using limit with groupBy.

My understanding is that this parameter limits the number of entries the Aggregate query will return, e.g. grouping step yielded 10 results, but we only want the first 5. The tests use `objectLimit` though and I'm not sure what the difference between the two is.

3. Since `PropertyOf<T>` already has the fallback for T is `undefined`, make it the default value for T in `AggregateGroupByOptions<T>`.

```ts
// The user can then omit the type parameter entirely
let regions: AggregateGroupByOptions = {groupBy: "location"}

// Instead of having to be explicit about it (I don't see that explicitness adds more expressiveness here)
let regions: AggregateGroupByOptions<undefined> = {groupBy: "location"}
```

Now that I've typed up the snippet above, I'm thinking the following: does allowing `undefined` make the user's code more error prone?
The guiding principle is that clients needn't do any such validation (it already being the server's responsibility), but in TS we get this for free (and with less type acrobatics) by forcing the user to pass the collection type.

src/collections/aggregate/index.ts
```diff
@@ -35,8 +34,21 @@ export type AggregateNearOptions<T, M> = AggregateBaseOptions<T, M> & {
...
+  vector?: number[];
+};
+
+export type AggregateGroupByHybridOptions<T, M> = AggregateHybridOptions<T, M> & {
+  groupBy: T extends undefined ? string : (keyof T & string) | GroupByAggregate<T>;
+};
+
 export type AggregateGroupByNearOptions<T, M> = AggregateNearOptions<T, M> & {
-  groupBy: (keyof T & string) | GroupByAggregate<T>;
+  groupBy: T extends undefined ? string : (keyof T & string) | GroupByAggregate<T>;
```
bevzzz, line 51: Two good candidates for `PropertyOf<T>` here 👥

src/collections/aggregate/index.ts
```diff
@@ -422,7 +451,7 @@ class AggregateManager<T> implements Aggregate<T> {
   base(
     metrics?: PropertiesMetrics<T>,
     filters?: FilterValue,
-    groupBy?: (keyof T & string) | GroupByAggregate<T>
+    groupBy?: (keyof T & string) | GroupByAggregate<T> | string
```
bevzzz, line 454: Is this `groupBy` meant to be of a different type from the ones above?

```ts
groupBy: T extends undefined ? string : (keyof T & string) | GroupByAggregate<T> // aka PropertyOf
```

If my understanding is correct, both should have the same effect.

src/collections/aggregate/index.ts
```diff
@@ -7,26 +7,25 @@ import { FilterValue } from '../filters/index.js';
...
 import { Aggregator } from '../../graphql/index.js';
-import { toBase64FromMedia } from '../../index.js';
+import { PrimitiveKeys, toBase64FromMedia } from '../../index.js';
+import { Bm25QueryProperty } from '../query/types.js';
 import { Serialize } from '../serialize/index.js';
 
 export type AggregateBaseOptions<T, M> = {
   filters?: FilterValue;
   returnMetrics?: M;
 };
```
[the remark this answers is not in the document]
tsmith023: Good call out! I think I originally had `T` being passed to `FilterValue` in a previous iteration of the types. I'll remove!

src/collections/aggregate/index.ts
tsmith023: 1. I like the suggestion of `PropertyOf`, I'll add it to this PR!
2. This looks like another instance of the type being defined in a previous iteration of the API design. As it stands, there's no reason, but previously the `groupby` field was used to perform conditional type mapping similar to the `.query` namespace. Sadly, changing this now would be a breaking change!
3. The `undefined` specification is for users that don't provide generics when getting/creating their collections so that the type system can return non-generic returns that are just `Record<string, WeaviateField>`. I keep `= undefined` restricted to the top level of the API so that internals don't have to deal with it (except for conditionally mapping it) and doesn't need to define the default all the time

src/collections/aggregate/index.ts
bevzzz: > both should have the same effect.

If the above's true and we agree on adding `PropertyOf`, could we use it here as well?

tsmith023: sure, will do!

bevzzz (commented):
OK, one more `PropertyOf` ⬆️

src/collections/aggregate/index.ts
```diff
@@ -7,36 +7,50 @@ import { FilterValue } from '../filters/index.js';
...
-  groupBy: (keyof T & string) | GroupByAggregate<T>;
+export type PropertyOf<T> = T extends undefined ? string : keyof T & string;
+
+export type AggregateGroupByOptions<T, M> = AggregateBaseOptions<M> & {
+  groupBy: PropertyOf<T> | GroupByAggregate<T>;
 };
 
 export type GroupByAggregate<T> = {
-  property: keyof T & string;
+  property: T extends undefined ? string : keyof T & string;
```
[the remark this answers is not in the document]
tsmith023: 🤦 made those changes on the wrong branch and then missed them when re-implementing on this one

Issue #222, opened 2024-11-04
Add support for updating generative and reranker moduleconfig


dirkkul:
for reference: https://github.com/weaviate/weaviate-python-client/pull/1376/files

PR #249, opened 2025-01-10
Improve generative/reranker config UX

tsmith023:
- Allow updating of generative and reranker configs dynamically
- Fix typing issues with generative and reranker configs

Closes https://github.com/weaviate/typescript-client/issues/222

### DIFF
--- a/src/collections/config/unit.test.ts
+++ b/src/collections/config/unit.test.ts
@@ -8,6 +8,8 @@ import { MergeWithExisting } from './classes';
 import { GenerativeCohereConfig, RerankerCohereConfig } from './types';
 
 describe('Unit testing of the MergeWithExisting class', () => {
+  const parse = (config: any) => JSON.parse(JSON.stringify(config));
+
   const invertedIndex: WeaviateInvertedIndexConfig = {
     bm25: {
       b: 0.8,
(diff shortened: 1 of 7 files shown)


src/collections/config/unit.test.ts
```diff
@@ -122,8 +124,22 @@ describe('Unit testing of the MergeWithExisting class', () => {
...
+      model: 'model',
+      maxTokensProperty: '5',
+      returnLikelihoodsProperty: 'likelihoods',
+      stopSequencesProperty: ['and'],
+      temperatureProperty: 5.2,
+    },
+    'reranker-cohere': {},
+  };
+
+  const parse = (config: any) => JSON.parse(JSON.stringify(config));
```
bevzzz, line 139: Wouldn't this be a no-op most of the time?

bevzzz:
Small nitpick in the tests, LGTM besides that

src/collections/config/unit.test.ts
tsmith023: Yes and no. It won't change the values but it acts as a copy of the config so that the defined configs can be reused between tests. This is because the `MergeWithExisting` static methods modify the passed `current` configs due to copy by reference

PR #252, opened 2025-01-13
feat: add the support of URLs for the `nearImage()` method

### LINKED ISSUE #232: Add URL string as supported input for `nearImage()`
Hey, we talked about this. 

As file paths are a valid parameter for `nearImage()` search, could we add support for URLs, too?

pacyL2K19:
# This PR

- Fix #232 

@tsmith023  can you have a look, please?

### DIFF
--- a/src/utils/base64.ts
+++ b/src/utils/base64.ts
@@ -23,7 +23,7 @@ const isFilePromise = (file: string | Buffer): Promise<boolean> =>
     });
   });
 
-const isUrl = (file: string | Buffer): boolean => {
+const isUrl = (file: string): file is string => {
   if (typeof file !== 'string') return false;
   try {
     const url = new URL(file);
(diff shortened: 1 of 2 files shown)


src/utils/base64.ts
```diff
@@ -22,6 +23,39 @@ const isFilePromise = (file: string | Buffer): Promise<boolean> =>
     });
   });
 
+const isUrl = (file: string | Buffer): boolean => {
```
tsmith023, line 26: Is the type required to be `string | Buffer` here? Can it not be just `string` because of the type narrowing [here](https://github.com/weaviate/typescript-client/pull/252/files#diff-a7864de1a55729a51dfe33cb5499284e6094ebc383e7bcd7f13dfe32d46d5bfaR72-R75) by `isBuffer`?

Can we also change it to be a true type guard so `: file is string` rather than `: boolean`? wdyt?

pacyL2K19: Makes 100% sense @tsmith023 
I just made the required changes, please have a look again

PR #263, opened 2025-02-10
Refactor `roles` namespace, adding `users`, to align with new Python API

tsmith023:
- Split `roles` with `users`
- make `actions` an array inside `permissions`
- rename several methods to be aligned with py client in a non-BC way
- only run roles/users tests for `>1.29` due to no BC

### DIFF
--- a/src/roles/index.ts
+++ b/src/roles/index.ts
@@ -14,15 +14,72 @@ import {
 import { Map } from './util.js';
 
 export interface Roles {
+  /**
+   * Retrieve all the roles in the system.
+   *
+   * @returns {Promise<Record<string, Role>>} A map of role names to their respective roles.
+   */
   listAll: () => Promise<Record<string, Role>>;
+  /**
+   * Retrieve a role by its name.
+   *
+   * @param {string} roleName The name of the role to retrieve.
+   * @returns {Promise<Role | null>} The role if it exists, or null if it does not.
+   */
   byName: (roleName: string) => Promise<Role | null>;
+  /**
+   * Retrieve the user IDs assigned to a role.
+   *
+   * @param {string} roleName The name of the role to retrieve the assigned user IDs for.
+   * @returns {Promise<string[]>} The user IDs assigned to the role.
+   */
   assignedUserIds: (roleName: string) => Promise<string[]>;
+  /**
+   * Delete a role by its name.
+   *
+   * @param {string} roleName The name of the role to delete.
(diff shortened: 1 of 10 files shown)


src/roles/index.ts
```diff
@@ -10,37 +10,29 @@ import {
...
+  assignedUserIds: (roleName: string) => Promise<string[]>;
   delete: (roleName: string) => Promise<void>;
   create: (roleName: string, permissions: PermissionsInput) => Promise<Role>;
-  assignToUser: (roleNames: string | string[], user: string) => Promise<void>;
   exists: (roleName: string) => Promise<boolean>;
-  revokeFromUser: (roleNames: string | string[], user: string) => Promise<void>;
   addPermissions: (roleName: string, permissions: PermissionsInput) => Promise<void>;
   removePermissions: (roleName: string, permissions: PermissionsInput) => Promise<void>;
-  hasPermission: (roleName: string, permission: Permission) => Promise<boolean>;
+  hasPermissions: (roleName: string, permission: Permission) => Promise<boolean>;
```
bevzzz, line 25:
```suggestion
  hasPermissions: (roleName: string, permission: Permission | Permission[]) => Promise<boolean>;
```

suggestion(blocking): I know TS has no problem with the implementation accepting a wider type than the interface declares, but this is probably truer to the method's name

src/roles/index.ts
```diff
@@ -10,37 +10,29 @@ import {
...
 } from './types.js';
 import { Map } from './util.js';
 
 export interface Roles {
   listAll: () => Promise<Record<string, Role>>;
-  ofCurrentUser: () => Promise<Record<string, Role>>;
   byName: (roleName: string) => Promise<Role | null>;
-  byUser: (user: string) => Promise<Record<string, Role>>;
-  assignedUsers: (roleName: string) => Promise<Record<string, User>>;
+  assignedUserIds: (roleName: string) => Promise<string[]>;
```
bevzzz, line 19: nit: I know this follows the naming in the Python client, but it feels slightly over-specified. Personally, I would expect the fact that "the strings returned by this method are user IDs" to be mentioned in the method doc, rather than be a part of the name.

bevzzz (approved):

src/roles/index.ts
tsmith023: I think we can do both! Also, good reminder that these methods need docstrings :D

PR #267, opened 2025-03-03
always pass embedding service headers

augustas1:
Always pass Embedding Service headers so it works also via OIDC auth


tsmith023 (approved):
Lovely typing changes too getting rid of the `request: any` 😁

PR #274, opened 2025-03-17
Add ability to define tenant-level perms for `data` and `tenants` perms

tsmith023:
Adds capability to define `tenant` and `data` permissions specific to certain tenants, e.g.:
```typescript
permissions.data({
  collection: "ColA",
  tenant: "tenant1",
  create: True
})
```

### DIFF
--- a/src/roles/index.ts
+++ b/src/roles/index.ts
@@ -121,6 +121,15 @@ const roles = (connection: ConnectionREST): Roles => {
 };
 
 export const permissions = {
+  /**
+   * Create a set of permissions specific to Weaviate's backup functionality.
+   *
+   * For all collections, provide the `collection` argument as `'*'`.
+   *
+   * @param {string | string[]} args.collection The collection or collections to create permissions for.
+   * @param {boolean} [args.manage] Whether to allow managing backups. Defaults to `false`.
+   * @returns {BackupsPermission[]} The permissions for the specified collections.
+   */
   backup: (args: { collection: string | string[]; manage?: boolean }): BackupsPermission[] => {
     const collections = Array.isArray(args.collection) ? args.collection : [args.collection];
     return collections.flatMap((collection) => {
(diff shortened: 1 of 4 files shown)


src/roles/integration.test.ts
```diff
@@ -1,12 +1,288 @@
...
+    },
+  },
+];
+
+const maybe = DbVersion.fromString(`v${process.env.WEAVIATE_VERSION!}`).isAtLeast(1, 29, 0)
   ? describe
   : describe.skip;
 
-only('Integration testing of the roles namespace', () => {
+maybe('Integration testing of the roles namespace', () => {
```
bevzzz, line 285: You could use [`describe.skipIf`](https://vitest.dev/api/#describe-skipif) if it looks more readable:

```ts
const version = DbVersion.fromString(...);
describe.skipIf(!version.isAtLeast(1, 29, 0), 'Integration testing ...' () => {
``` 

There's a good alias for it too: `suite.skipIf`

src/roles/index.ts
```diff
@@ -208,32 +280,62 @@ export const permissions = {
...
+   *
+   * @param {string | string[]} args.user The user or users to create permissions for.
+   * @param {boolean} [args.assignAndRevoke] Whether to allow assigning and revoking users. Defaults to `false`.
+   * @param {boolean} [args.read] Whether to allow reading users. Defaults to `false`.
+   * @returns {UsersPermission[]} The permissions for the specified users.
+   */
   users: (args: {
     user: string | string[];
-    assign_and_revoke?: boolean;
+    assignAndRevoke?: boolean;
```
bevzzz, line 332: Would this not be a breaking change?

src/roles/index.ts
```diff
@@ -151,24 +176,54 @@ export const permissions = {
...
     });
   },
+  /**
+   * Create a set of permissions specific to any operations involving objects within collections and tenants.
+   *
+   * For all collections, provide the `collection` argument as `'*'`.
+   * For all tenants, provide the `tenant` argument as `'*'`.
+   *
+   * Providing arrays of collections and tenants will create permissions for each combination of collection and tenant.
+   * E.g., `data({ collection: ['A', 'B'], tenant: ['X', 'Y'] })` will create permissions for tenants `X` and `Y` in both collections `A` and `B`.
```
bevzzz, line 186: suggestion: perhaps this notice would be better placed in to top-most `export permissions {...}` docs, since the idea applies to all permissions which have more than 1 non-action parameter.

src/roles/index.ts
```diff
@@ -129,11 +138,27 @@ export const permissions = {
...
+   *
+   * @param {string | string[]} args.collection The collection or collections to create permissions for.
+   * @param {boolean} [args.create_collection] Whether to allow creating collections. Defaults to `false`.
+   * @param {boolean} [args.read_config] Whether to allow reading collection configurations. Defaults to `false`.
+   * @param {boolean} [args.update_config] Whether to allow updating collection configurations. Defaults to `false`.
+   * @param {boolean} [args.delete_collection] Whether to allow deleting collections. Defaults to `false`.
+   */
   collections: (args: {
     collection: string | string[];
     create_collection?: boolean;
```
bevzzz, line 164: (Github wouldn't let me comment an "originally-collapsed block")

note: while not confusing, I noticed the "actions" here specify _collection/_config, whereas other permissions simply have create/read/update/delete. Do you think this distinction is necessary?

bevzzz (approved):

src/roles/integration.test.ts
tsmith023: This doesn't appear available in the Jest version of the client. I found [this](https://github.com/jestjs/jest/issues/3652#issuecomment-303964364) but it looks like it was never merged. Do you have a source for `describe.skipIf`?

src/roles/index.ts
tsmith023: That's true and I'd agree if there were more than 2 permissions with more than one non-action param. But I think users would be much less likely to read the top-level documentation on this than the docstring directly next to its usage, wdyt?

src/roles/integration.test.ts
bevzzz: Right, I should've checked the version we're using rn 😅

src/roles/index.ts
bevzzz: fair point, I was thinking that too

bevzzz (approved):

PR #278, opened 2025-03-31
feat: support dynamic user management in Weaviate 1.30

bevzzz:
This PR adds `user.db` and `user.oidc` namespaces with methods for creating users dynamically and managing their permissions.


src/roles/util.ts
```diff
@@ -160,7 +187,11 @@ class PermissionsMapping {
   public static use = (role: WeaviateRole) => new PermissionsMapping(role);
 
   public map = (): Role => {
-    this.role.permissions.forEach(this.permissionFromWeaviate);
+    // If truncated roles are requested (?includeFullRoles=false),
+    // role.permissions are not present.
+    if (this.role.permissions !== null) {
+      this.role.permissions.forEach(this.permissionFromWeaviate);
+    }
```
tsmith023, line 194: I think, because of how the OpenAPI schema is defined, this clause won't fire. I think you should add a test of `.getAssignedRoles` with `includePermissions: true` and see what happens. But, given the Python implementation, I think you will ned to pas `inlcudePermissions` here to the `map` function and use whether that is true or not to parse the `this.role.permissions`. Either way, I think it is returned as an empty list even if `includePermissions: false`

src/users/index.ts
```diff
@@ -18,36 +46,139 @@ export interface Users {
...
+   * Retrieve the roles assigned to a user.
    *
-   * @param {string | string[]} roleNames The name or names of the roles to assign.
-   * @param {string} userId The ID of the user to assign the roles to.
-   * @returns {Promise<void>} A promise that resolves when the roles are assigned.
+   * @param {string} userId The ID of the user to retrieve the assigned roles for.
+   * @returns {Promise<Record<string, Role>>} A map of role names to their respective roles.
    */
-  assignRoles: (roleNames: string | string[], userId: string) => Promise<void>;
+  getAssignedRoles: (userId: string, includePermissions?: boolean) => Promise<Record<string, Role>>;
```
tsmith023, line 62: Despite this being the only optional parameter for now in this method, it could be extended at some point in the future. Can we introduce a type here, say `type GetAssignedRolesOptions = { includePermissions?: boolean }`, and have the second argument here be `opts?: GetAssignedRolesOptions`?

That way usage would be more inline with the rest of the codebase, e.g.
```typescript
.getAssignedRoles("id", { includePermissions: true })
```
and this would be easily extendable if we need to add other types of includes depending on the future complexity of the role object

src/users/index.ts
```diff
@@ -18,36 +46,139 @@ export interface Users {
...
+   * Retrieve the roles assigned to a user.
    *
-   * @param {string | string[]} roleNames The name or names of the roles to revoke.
-   * @param {string} userId The ID of the user to revoke the roles from.
-   * @returns {Promise<void>} A promise that resolves when the roles are revoked.
+   * @param {string} userId The ID of the user to retrieve the assigned roles for.
+   * @returns {Promise<Record<string, Role>>} A map of role names to their respective roles.
    */
-  revokeRoles: (roleNames: string | string[], userId: string) => Promise<void>;
+  getAssignedRoles: (userId: string, includePermissions?: boolean) => Promise<Record<string, Role>>;
```
tsmith023, line 81: Same here for the `GetAssignedRolesOptions` type

src/users/index.ts
```diff
@@ -18,36 +46,139 @@ export interface Users {
...
+const db = (connection: ConnectionREST): DBUsers => {
+  const ns = namespacedUsers(connection);
+
+  const allowCode = (code: number): ((reason: any) => boolean) => {
+    return (reason) => reason.code !== undefined && reason.code === code;
+  };
+
+  type APIKeyResponse = { apikey: string };
+  return {
+    getAssignedRoles: (userId: string, includePermissions?: boolean) =>
```
tsmith023, line 107: Same here for the `GetAssignedRolesOptions` type

src/users/index.ts
```diff
@@ -18,36 +46,139 @@ export interface Users {
...
+        .catch(allowCode(409)),
+    byName: (userId: string) => connection.get<WeaviateDBUser>(`/users/db/${userId}`, true).then(Map.dbUser),
+    listAll: () => connection.get<WeaviateDBUser[]>('/users/db', true).then(Map.dbUsers),
+  };
+};
+
+const oidc = (connection: ConnectionREST): OIDCUsers => {
+  const ns = namespacedUsers(connection);
+  return {
+    getAssignedRoles: (userId: string, includePermissions?: boolean) =>
```
tsmith023, line 139: Same here for the `GetAssignedRolesOptions` type

src/users/index.ts
```diff
@@ -18,36 +46,139 @@ export interface Users {
...
+    revokeRoles: (roleNames: string | string[], userId: string) => ns.revokeRoles(roleNames, userId, 'oidc'),
+  };
+};
+
+/** Internal interface for operations that MAY accept a 'db'/'oidc' namespace. */
+interface NamespacedUsers {
+  getAssignedRoles: (
+    userType: UserTypeInternal,
+    userId: string,
+    includePermissions?: boolean
```
tsmith023, line 151: Same here for the `GetAssignedRolesOptions` type

src/users/index.ts
```diff
@@ -18,36 +46,139 @@ export interface Users {
...
+  const ns = namespacedUsers(connection);
+  return {
+    assignRoles: (roleNames: string | string[], userId: string) => ns.assignRoles(roleNames, userId),
+    revokeRoles: (roleNames: string | string[], userId: string) => ns.revokeRoles(roleNames, userId),
+  };
+};
+
+const namespacedUsers = (connection: ConnectionREST): NamespacedUsers => {
+  return {
+    getAssignedRoles: (userType: UserTypeInternal, userId: string, includePermissions?: boolean) =>
```
tsmith023, line 167: Same here for the `GetAssignedRolesOptions` type

src/users/index.ts
```diff
@@ -18,36 +46,139 @@ export interface Users {
...
+      connection
+        .get<WeaviateRole[]>(
+          `/authz/users/${userId}/roles/${userType}${includePermissions ? '?&includeFullRoles=true' : ''}`
+        )
+        .then(Map.roles),
+    assignRoles: (roleNames: string | string[], userId: string, userType?: UserTypeInternal) =>
```
tsmith023, line 173: wdyt about having an optional object type here for `userType`? Is it as useful as the other instance?

src/users/index.ts
```diff
@@ -18,36 +46,139 @@ export interface Users {
...
       connection.postEmpty(`/authz/users/${userId}/assign`, {
         roles: Array.isArray(roleNames) ? roleNames : [roleNames],
+        userType: userType,
       }),
-    revokeRoles: (roleNames: string | string[], userId: string) =>
+    revokeRoles: (roleNames: string | string[], userId: string, userType?: UserTypeInternal) =>
```
tsmith023, line 178: wdyt about having an optional object type here for `userType`? Is it as useful as the other instance?

test/version.ts
```diff
@@ -0,0 +1,7 @@
+import { DbVersion } from '../src/utils/dbVersion';
+
+const version = DbVersion.fromString(`v${process.env.WEAVIATE_VERSION!}`);
+
+/** Run the suite / test only for Weaviate version above this. */
+export const requireAtLeast = (...semver: [...Parameters<DbVersion['isAtLeast']>]) =>
+  version.isAtLeast(...semver) ? describe : describe.skip;
```
tsmith023, line 7: Very happy with this abstraction! Hopefully we get a chance to move the other instances of `maybe` to use this

src/users/index.ts
bevzzz: > That way usage would be more inline with the rest of the codebase

Do I gather correctly that our convention then is to group optional arguments in one optional object?
I'll update all the call sites then

test/version.ts
bevzzz: Thank you!
I've been updating all the test files I needed to modify as I went, we'll soon update them all.

I was thinking about returning an object `{ describe, test }` at first, but so far haven't had the need to do anything like this:

```ts
requireAtLeast(1, 28, 4).describe('desc', () => {
  requireAtLeast(1, 29, 2).test('', () => {})
  requireAtLeast(1, 29, 5).test('', () => {})
})
```

Grouping all related tests under one "require" block seems sufficient.

src/roles/util.ts
bevzzz: Let me check again -- I am rather positive that I've only added this clause because a test I wrote for `userAssignments` was initially failing.

bevzzz: Getting this error if the if-clause is removed:

(in the test context)
```sh
Rejected to value: [TypeError: Cannot read properties of null (reading 'forEach')]
```

I'll add that test to the suite to illustrate.

tsmith023: Okay, very interesting! Thanks for verifying 😁

src/users/index.ts
tsmith023: Yes, that's correct! E.g., for searching:
```typescript
res = await collection.query.near_text("A query", {
  limit: 5,
  returnMetadata: ['distance'],
})
```

src/users/index.ts
```diff
@@ -18,35 +46,151 @@ export interface Users {
...
+    revokeRoles: (roleNames: string | string[], userId: string) => base.revokeRoles(roleNames, userId),
+    db: db(connection),
+    oidc: oidc(connection),
+  };
+};
+
+const db = (connection: ConnectionREST): DBUsers => {
+  const ns = namespacedUsers(connection);
+
+  const allowCode = (code: number): ((reason: any) => boolean) => {
```
tsmith023, line 105: I like this method but the `reason: any` here is a concern type-safety-wise 🤔

I see that this is used in `.catch` methods so can we change it to perform a type guard inside `allowCode` for when the error thrown is a known unexpected status code? E.g.
```typescript
const allowCode = (code: number): ((reason: any) => boolean) => (reason) => {
  if (reason instanceof WeaviateUnexpectedStatusCodeError) {
    return reason.code === code;
  };
  throw reason;
}
```
This way, if the thrown error is something else (somehow) then that error is still propagated up to the user appropriately. wdyt?

src/users/index.ts
```diff
@@ -18,35 +46,151 @@ export interface Users {
...
+  rotateKey: (userId: string) => Promise<string>;
+  activate: (userId: string) => Promise<boolean>;
+  deactivate: (userId: string) => Promise<boolean>;
+  byName: (userId: string) => Promise<UserDB>;
+  listAll: () => Promise<UserDB[]>;
+}
+
+type GetAssingedRolesOptions = {
+  includePermissions?: boolean;
+};
```
tsmith023, line 75: Can we do the following here?
- Move this to `src/roles/types.ts`
- Export the type from that file
- Fix the `Assinged` typo to `Assigned`

Cheers! 😁

src/roles/util.ts
```diff
@@ -123,20 +130,38 @@ export class Map {
...
   static users = (users: string[]): Record<string, User> =>
-    users.reduce((acc, user) => {
-      acc[user] = { id: user };
-      return acc;
-    }, {} as Record<string, User>);
+    users.reduce(
+      (acc, user) => ({
+        ...acc,
+        [user]: { id: user },
+      }),
```
tsmith023, line 146: Nice syntactical change!

src/users/index.ts
```diff
@@ -18,35 +46,151 @@ export interface Users {
...
+const baseUsers = (connection: ConnectionREST): UsersBase => {
+  const ns = namespacedUsers(connection);
+  return {
+    assignRoles: (roleNames: string | string[], userId: string) => ns.assignRoles(roleNames, userId),
+    revokeRoles: (roleNames: string | string[], userId: string) => ns.revokeRoles(roleNames, userId),
+  };
+};
+
+/** Optional arguments to /assign and /revoke endpoints. */
+type AssignRevokeOptions = { userType?: UserTypeInternal };
```
tsmith023, line 174: Likewise here for moving to `types.ts` file and exporting from there

src/users/index.ts
bevzzz: Makes sense to me 👍

src/roles/util.ts
bevzzz: Saw it on SO and thought it looks [idiomatic](https://media2.giphy.com/media/v1.Y2lkPTc5MGI3NjExN2h1b3I2Yng5YnRnN2t6cWJrdnFzNjZybnZoYWxmajByNTQxd2x2dyZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/3oriePYGXYh3bPg0Pm/giphy.gif) :)

src/users/index.ts
```diff
@@ -18,35 +46,151 @@ export interface Users {
...
+      ns.assignRoles(roleNames, userId, { userType: 'db' }),
+    revokeRoles: (roleNames: string | string[], userId: string) =>
+      ns.revokeRoles(roleNames, userId, { userType: 'db' }),
+
+    create: (userId: string) =>
+      connection.postNoBody<APIKeyResponse>(`/users/db/${userId}`).then((resp) => resp.apikey),
+    delete: (userId: string) =>
+      connection
+        .delete(`/users/db/${userId}`, null)
+        .then(() => true)
```
mpartipilo, line 123: Does work when the returned code is 204? That's what the server returns.

src/users/index.ts
bevzzz: Totes 💯

src/users/index.ts
bevzzz: `connection` will only throw errors for codes >= 400, so we'll be returning `true` for any 2** code here

https://github.com/weaviate/typescript-client/blob/b9bdb0c63922a91a2f5ab04f2348327be24dc69c/src/connection/http.ts#L373-L377

src/users/index.ts
```diff
@@ -18,35 +47,202 @@ export interface Users {
...
+   */
+  activate: (userId: string) => Promise<boolean>;
+
+  /**
+   * Deactivate 'db_user' user.
+   *
+   * @param {string} userId The ID of the user to deactivate.
+   * @returns {Promise<boolean>} `true` if the user has been successfully deactivated.
+   */
+  deactivate: (userId: string) => Promise<boolean>;
```
dirkkul, line 104: There is a revoke_key option that is missing (I also forgot it in the python client)

tsmith023: `revokeKey` in TS plz 😜

tsmith023 (approved):
Thanks for fixing the mock batching test issues!

PR #279, opened 2025-04-01
Improvements on typing

tsmith023:
- Fixes bad typing of `GenerativeConfigRuntime`
- adds mocks for this usage
- renames `generativeConfigRuntime` to `generativeParameters` to match Python naming

### DIFF
--- a/src/collections/types/generate.ts
+++ b/src/collections/types/generate.ts
@@ -122,18 +122,18 @@ export type GroupedTask<T> = {
 type omitFields = 'images' | 'imageProperties';
 
 export type GenerativeConfigRuntime =
-  | ModuleConfig<'generative-anthropic', GenerativeConfigRuntimeType<'generative-anthropic'>>
-  | ModuleConfig<'generative-anyscale', GenerativeConfigRuntimeType<'generative-anyscale'>>
-  | ModuleConfig<'generative-aws', GenerativeConfigRuntimeType<'generative-aws'>>
+  | ModuleConfig<'generative-anthropic', GenerativeConfigRuntimeType<'generative-anthropic'> | undefined>
+  | ModuleConfig<'generative-anyscale', GenerativeConfigRuntimeType<'generative-anyscale'> | undefined>
+  | ModuleConfig<'generative-aws', GenerativeConfigRuntimeType<'generative-aws'> | undefined>
   | ModuleConfig<'generative-azure-openai', GenerativeConfigRuntimeType<'generative-azure-openai'>>
-  | ModuleConfig<'generative-cohere', GenerativeConfigRuntimeType<'generative-cohere'>>
-  | ModuleConfig<'generative-databricks', GenerativeConfigRuntimeType<'generative-databricks'>>
-  | ModuleConfig<'generative-dummy', GenerativeConfigRuntimeType<'generative-dummy'>>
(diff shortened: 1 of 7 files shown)


src/collections/types/generate.ts
```diff
@@ -122,18 +122,18 @@ export type GroupedTask<T> = {
...
-  | ModuleConfig<'generative-nvidia', GenerativeConfigRuntimeType<'generative-nvidia'>>
-  | ModuleConfig<'generative-ollama', GenerativeConfigRuntimeType<'generative-ollama'>>
+  | ModuleConfig<'generative-cohere', GenerativeConfigRuntimeType<'generative-cohere'> | undefined>
+  | ModuleConfig<'generative-databricks', GenerativeConfigRuntimeType<'generative-databricks'> | undefined>
+  | ModuleConfig<'generative-dummy', GenerativeConfigRuntimeType<'generative-dummy'> | undefined>
+  | ModuleConfig<'generative-friendliai', GenerativeConfigRuntimeType<'generative-friendliai'> | undefined>
+  | ModuleConfig<'generative-google', GenerativeConfigRuntimeType<'generative-google'> | undefined>
+  | ModuleConfig<'generative-mistral', GenerativeConfigRuntimeType<'generative-mistral'> | undefined>
+  | ModuleConfig<'generative-nvidia', GenerativeConfigRuntimeType<'generative-nvidia'> | undefined>
+  | ModuleConfig<'generative-ollama', GenerativeConfigRuntimeType<'generative-ollama'> | undefined>
```
bevzzz, line 136: suggestion: make `ModuleConfig.config` optional (`config?: C`), then have individual ModuleConfig variants make it required explicitly?

https://github.com/weaviate/typescript-client/blob/f3ab9f89e62d1ec46f23e2a7c9788cc25de70df0/src/collections/config/types/index.ts#L17-L20

Idk, maybe it's less flexible, but I just noticed the repeated `| undefined` and thought about this.

Example [TS Playground](https://www.typescriptlang.org/play/?ssl=11&ssc=54&pln=10&pc=1#code/KYDwDg9gTgLgBDAnmYcCyEAmBXANsAYQgDsAzASwHMAeAOQBo4C4BeObYzYC44TAPlZwA3gCg4cYgEMAtsABccWgG5xcAMYkKlAPyKCyuAHojcagFpzcKbgDuUxAGc4AIghgY5EjZeiAvqqiSChwAOLAvFBSngBuqGziGDj4RGRU1ADkAIIZjMTYMgBGwFDGAFRwAD7snNzkvJhwZUb8otVJeIRa6RkAQrmSBcWlRhXVHFw8fE0tbXAASsAAjtjkUHzUHSndNBkEA-lFJfz8gZrEjvBSiuGR0eRxQsLScorZA+faihN1DQGi50ucEKNwiJXujzYz1kCjgfQ+O2+tSmmH+gPg6lBd1i8REL1hewRaUoigAzAFjKYLFZPlRJBB4O5PN5cHBREA)

bevzzz (approved):

src/collections/types/generate.ts
tsmith023: This one's because of annoying aspect of the design decision here that I want the type `C` to be inferred as `undefined` if it is. If you set `config?: C` then doing `config: undefined` won't infer `C` as `undefined` but instead as `unknown`. Agreed that the `| undefined` is cumbersome here but I think it's necessary because of the above

PR #285, opened 2025-04-09
fix: revert breaking change and add deprecation notices

bevzzz:
Follow-up for #278 

Unfortunately, the initial PR adding dynamic user management had a couple of blunders, which I'm addressing in this PR.

- ‼️ **Bring back `roles.assignedUserIds`** -- I'd got a false impression that this method was removed in the Python client and [replaced it with _userAssignments_](https://github.com/weaviate/typescript-client/pull/278/commits/d61c30fbf3ab0ea18225d86f734ba6287d6f7fdd) rather than deprecating it.
- Return `false` on HTTP 409 Conflict from /activate and /deactivate endpoints.
- Add deprecation notices to `users.assignRoles`, `users.revokeRoles`, and `roles.assignedUserIds`

### DIFF
--- a/src/users/index.ts
+++ b/src/users/index.ts
@@ -34,6 +34,11 @@ interface UsersBase {
 }
 
 export interface Users extends UsersBase {
+  /** @deprecated: Use `users.db.assignRoles` or `users.oidc.assignRoles` instead. */
+  assignRoles: (roleNames: string | string[], userId: string) => Promise<void>;
+  /** @deprecated: Use `users.db.revokeRoles` or `users.oidc.revokeRoles` instead. */
+  revokeRoles: (roleNames: string | string[], userId: string) => Promise<void>;
+
   /**
    * Retrieve the information relevant to the currently authenticated user.
    *
--- a/src/users/index.ts
+++ b/src/users/index.ts
@@ -147,11 +147,12 @@ const users = (connection: ConnectionREST): Users => {
 const db = (connection: ConnectionREST): DBUsers => {
   const ns = namespacedUsers(connection);
 
-  /** expectCode returns true if the error contained an expected status code. */
+  /** expectCode returns false if the contained WeaviateUnexpectedStatusCodeError
+   * has an known error code and rethrows the error otherwise. */
   const expectCode = (code: number): ((_: any) => boolean) => {
     return (error) => {
-      if (error instanceof WeaviateUnexpectedStatusCodeError) {
-        return error.code === code;
+      if (error instanceof WeaviateUnexpectedStatusCodeError && error.code === code) {
+        return false;
       }
       throw error;
     };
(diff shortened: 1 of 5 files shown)


src/users/integration.test.ts
```diff
@@ -79,13 +79,13 @@ requireAtLeast(
       await expectDave().toHaveProperty('active', true);
 
       // Second activation is a no-op
-      await expect(client.users.db.activate('dynamic-dave')).resolves.toEqual(true);
+      await expect(client.users.db.activate('dynamic-dave')).resolves.toEqual(false);
 
       await client.users.db.deactivate('dynamic-dave');
```
dirkkul, line 84: can you add `.resolves.toEqual(true)`

src/users/integration.test.ts
```diff
@@ -79,13 +79,13 @@ requireAtLeast(
...
       await expectDave().toHaveProperty('active', false);
 
       // Second deactivation is a no-op
-      await expect(client.users.db.deactivate('dynamic-dave', { revokeKey: true })).resolves.toEqual(true);
+      await expect(client.users.db.deactivate('dynamic-dave', { revokeKey: true })).resolves.toEqual(false);
 
       await client.users.db.delete('dynamic-dave');
```
dirkkul, line 90: can you also double check the return code for delete? It should also return `true` if the user was deleted and `false` if the user did not exist at this time

src/users/index.ts
```diff
@@ -34,6 +34,11 @@ interface UsersBase {
 }
 
 export interface Users extends UsersBase {
+  /** @deprecated: Use `users.db.assignRoles` or `users.oidc.assignRoles` instead. */
+  assignRoles: (roleNames: string | string[], userId: string) => Promise<void>;
+  /** @deprecated: Use `users.db.revokeRoles` or `users.oidc.revokeRoles` instead. */
+  revokeRoles: (roleNames: string | string[], userId: string) => Promise<void>;
```
dirkkul, line 40: getAssignedRoles is also deprecated

bevzzz: addressed in https://github.com/weaviate/typescript-client/pull/285/commits/15a56a48fa0493d2985d8242844c07b598be3891

src/users/integration.test.ts
bevzzz: addressed in https://github.com/weaviate/typescript-client/pull/285/commits/7d9310ab0621162d4bde50c1c09bbf7c222ef58c

src/users/integration.test.ts
bevzzz: addressed in https://github.com/weaviate/typescript-client/pull/285/commits/7d9310ab0621162d4bde50c1c09bbf7c222ef58c

tsmith023 (approved):

PR #288, opened 2025-04-15
feat(v1.30.1): retrieve users' last login time and preview API keys

bevzzz:
### What is changed?
This PR adds optional `includeLastUsedTime` parameter to `users.db.byName` and `users.db.listAll` methods, which controls if the last login time will be retrieved for the users.

Additionally, `UserDB` model has 2 new fields:
- `createdAt`: returned for all users (provided server is `v1.30.1` and above)
- `apiKeyFirstLetters`: returned whenever request is made by the cluster admin (intended to be displayed in the WCS Console)

Updated `openapi.ts` schema spec.

### TODO
- [x] Run tests with `v1.30.1`

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -125,7 +125,7 @@ jobs:
           registry-url: 'https://registry.npmjs.org'
       - run: npm ci
       - run: npm run build
-      - run: npm publish --tag alpha
+      - run: npm publish
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_AUTOMATION_TOKEN }}
       - run: npm run docs
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -13,7 +13,7 @@ env:
   WEAVIATE_127: 1.27.15
   WEAVIATE_128: 1.28.11
   WEAVIATE_129: 1.29.1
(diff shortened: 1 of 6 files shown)


.github/workflows/main.yaml
```diff
@@ -125,7 +125,7 @@ jobs:
           registry-url: 'https://registry.npmjs.org'
       - run: npm ci
       - run: npm run build
-      - run: npm publish
+      - run: npm publish --tag alpha
```
tsmith023, line 128: Make sure to change this back to `npm publish` before merging into `main`!

src/users/types.ts
```diff
@@ -11,6 +11,10 @@ export type UserDB = {
   id: string;
   roleNames: string[];
   active: boolean;
+
+  createdAt?: Date;
+  lastUsedAt?: Date;
```
tsmith023, line 16: Will we drop the `At` here or leave it?

tsmith023 (changes requested):

src/users/types.ts
bevzzz: Leave it 👌 

see https://weaviate-org.slack.com/archives/C064AL0Q6F2/p1746527397954569?thread_ts=1745570462.670529&cid=C064AL0Q6F2

tsmith023 (approved):

PR #307, opened 2025-06-20
Feat/introduce web client

### LINKED ISSUE #145: weaviate-client fails to bundle on Cloudflare Workers due to http2 dependency
## Description

I am trying to use the `weaviate-client` library in a Cloudflare Worker, but I am encountering issues related to the `http2` module. The error message suggests that the `http2` package is built into Node.js and is not available in the Cloudflare Workers environment.

## Context

I have added the following configuration to my `wrangler.toml` file to enable Node.js compatibility:

```toml
compatibility_date = "2023-12-01"
node_compat = true
```

However, even with this configuration, I am still facing the `http2` related error.

## Steps to Reproduce

1. Set up a new Cloudflare Worker project.
2. Install the `weaviate-client` library using npm or yarn.
3. Configure `wrangler.toml` with the appropriate `compatibility_date` and `node_compat` settings.
4. Write code that uses the `weaviate-client` library to make requests to a Weaviate instance.
5. Attempt to run the Worker using `wrangler dev`.

## Expected Behavior

The Cloudflare Worker should be able to use the `weaviate-client` library to make requests to a Weaviate instance without any issues related to the `http2` module.

## Actual Behavior

The Worker fails to execute due to the missing `http2` module, which is not available in the Cloudflare Workers environment.

```
✘ [ERROR] Could not resolve "http2"

    ../../node_modules/.pnpm/@grpc+grpc-js@1.9.14/node_modules/@grpc/grpc-js/build/src/transport.js:20:22:
      20 │ const http2 = require("http2");
         ╵                       ~~~~~~~

  The package "http2" wasn't found on the file system but is built into node. Are you trying to bundle for node? You can use "platform: 'node'" to do that, which will remove this error.
```

## Possible Solution

One potential solution is to use the native `fetch` API provided by Cloudflare Workers instead of relying on the `http2` module used by the `weaviate-client` library.

## Additional Information

- I'm using Weaviate Cloud
- `weaviate-client` library version: `3.0.0-rc.3`

tsmith023:
- Refactor into a monorepo composed of four packages:
  - `@weaviate/core`: shared functionality
  - `@weaviate/node`: Node-client that uses `nice-grpc` and `fs`
  - `@weaviate/web`: Browser-client that uses `nice-grpc-web` and `FileReader`
  - `@weaviate/test`: Tests the public APIs of the core/node pkgs
- Use `pnpm` for managing the monorepo
- Use dependency injection to pipe node/browser deps into the common functionality of the client
  - This will allow a `@weaviate/deno` package too with its own deno-specific libs injected in to close
  https://github.com/weaviate/typescript-client/issues/309
- Depends on core: https://github.com/weaviate/weaviate/pull/12024
- Closes https://github.com/weaviate/typescript-client/issues/145


packages/core/src/connection/auth.ts
```diff
@@ -197,22 +196,6 @@ class UserPasswordAuthenticator implements OidcAuthFlow {
...
-    ) {
-      throw new Error('grant_type password not supported');
-    }
-    if (this.openidConfig.provider.token_endpoint.includes('https://login.microsoftonline.com')) {
-      throw new Error(
-        'microsoft/azure recommends to avoid authentication using ' +
-          'username and password, so this method is not supported by this client'
-      );
-    }
-    this.openidConfig.scopes.push('offline_access');
```
bevzzz: issue: this is a breaking change, client should continue to add `offline_access` scope.

Sneaky having that at the end of something called "validate config" 😅
I missed it too while cherry-picking the commit but good thing the tests caught it.

packages/core/src/collection/index.ts
```diff
@@ -111,28 +112,31 @@ const isString = (value: any): value is string => typeof value === 'string';
 const capitalizeCollectionName = <N extends string>(name: N): N =>
   (name.charAt(0).toUpperCase() + name.slice(1)) as N;
 
-const collection = <T, N, V>(
+const collection = <T, N, V, M>(
```
bevzzz, line 115: nit/suggestion: if renaming generic parameters and formatting changes were their own PR(s) it'd be easier to focus on the key areas of the diff in this one

packages/core/src/connection/grpc.ts
```diff
@@ -43,76 +29,66 @@ export default class ConnectionGRPC extends ConnectionGQL {
   private grpc: GrpcClient;
   public grpcMaxMessageLength: number;
   private params: GrpcConnectionParams & { grpcMaxMessageLength: number };
+  private transports: Transports;
 
-  private constructor(params: GrpcConnectionParams & { grpcMaxMessageLength: number }) {
+  private constructor(
+    transports: Transports,
+    params: GrpcConnectionParams & { grpcMaxMessageLength: number }
```
bevzzz, line 36: question: Can `grpcMaxMessageLength` be included in the `GrpcConnectionParams` definition above on line 19?

packages/core/src/connection/grpc.ts
```diff
@@ -43,76 +29,66 @@ export default class ConnectionGRPC extends ConnectionGQL {
...
-      };
-    }
-    return Promise.all([
-      ConnectionGRPC.connect(
-        params,
+    let grpcMaxMessageLength = MAX_GRPC_MESSAGE_LENGTH;
+    if (!params.skipInitChecks) {
+      grpcMaxMessageLength = await Promise.all([
         (rest.get('/meta', true) as Promise<Meta>).then(
           (res: Meta) => res.grpcMaxMessageSize || MAX_GRPC_MESSAGE_LENGTH
```
bevzzz, line 53: issue: it feels wrong that `skipInitChecks` can also make the client ignore the server's gRPC message size limit; perhaps it should only control compatibility checks?

bevzzz: I understand that calling `/meta` with `skipInitChecks: false` will effectively perform _an_ init check despite their having been disabled. At the same time, respecting the server's transport configuration seems more important (?)

WDYT?

packages/core/src/connection/grpc.ts
```diff
@@ -222,26 +194,26 @@ export const grpcClient = (config: GrpcConnectionParams & { grpcMaxMessageLength
...
-        client,
+        transports.weaviate,
         collection,
         getMetadataWithEmbeddingServiceAuth(config, bearerToken),
         config.timeout?.insert || 90,
         consistencyLevel,
         tenant
       ),
-    close: () => channel.close(),
+    close: () => {},
```
bevzzz, line 204: question: why is `close` a no-op now?

packages/core/src/connection/grpc.ts
```diff
@@ -222,26 +194,26 @@ export const grpcClient = (config: GrpcConnectionParams & { grpcMaxMessageLength
...
     health: () => {
       const controller = new AbortController();
       const timeoutId = setTimeout(() => controller.abort(), (config.timeout?.init || 2) * 1000);
-      return health
+      return transports.health
         .check(
           { service: '/grpc.health.v1.Health/Check' },
```
bevzzz, line 210: It's quite surprising that `HealthClient` from Google's gRPC package doesn't have the name of the service "baked in", [what](https://www.theguardian.com/notesandqueries/query/0,,-203057,00.html)?

packages/core/src/index.ts
```diff
@@ -105,12 +97,12 @@ export type ClientParams = {
   skipInitChecks?: boolean;
 };
 
-export interface WeaviateClient {
+export interface IWeaviateClient<TMedia = any> {
   alias: Aliases;
   backup: Backup;
-  batch: Batch;
+  // batch: Batch;
```
bevzzz, line 103: question: Is this commented out temporarily?

packages/core/src/index.ts
```diff
@@ -18,33 +15,28 @@ import {
...
 import { Meta } from './openapi/types.js';
 import roles, { Roles, permissions } from './roles/index.js';
 import { DbVersion } from './utils/dbVersion.js';
+import { Backend, BackupCompressionLevel, BackupStatus } from './v2/backup/index.js';
+import MetaGetter from './v2/misc/metaGetter.js';
 
-import { Agent as HttpAgent } from 'http';
-import { Agent as HttpsAgent } from 'https';
-import { LiveChecker, OpenidConfigurationGetter, ReadyChecker } from './misc/index.js';
+import type { Agent } from 'http';
```
bevzzz, line 26: nit: I think `Agent as HttpAgent` was rather helpful

packages/core/src/index.ts
```diff
@@ -137,106 +129,60 @@ const cleanHost = (host: string, protocol: 'rest' | 'grpc') => {
...
+export type Context<TMedia> = {
+  transportsMaker: TransportsMaker;
+  toBase64FromMedia: ToBase64FromMedia<TMedia>;
+  /**
+   * Creates the HTTP(S) keep-alive agent for the REST/GraphQL connection. Supplied by the Node shim
+   * (`@weaviate/node`); omitted by the browser shim (`@weaviate/web`) so no Node `http`/`https`
+   * builtins are pulled into the browser bundle. When undefined, `fetch` runs without a custom agent.
+   */
+  agentMaker?: (secure: boolean) => Agent | undefined;
+};
```
bevzzz, line 141: issue: It jumps at me that `TMedia` has to be threaded through most of the lib even though it only seems to matter for "near media" queries, which is a teeny-tiny subset of the client APIs.

suggestion: I think dropping `ToBase64FromMedia` from the Context's / client's definition and leaving it up to the user to pick the right base64 util and convert their `Blob ^ Buffer` to string will prove a much simpler approach.

As an example, think of _an_ HTTP client (axios, fetch, etc): it doesn't have a generic parameter describing what the _source type_ of the payload is. It expects a `string` ( even `{}`-object` == any!) and it's up to the user to serialize their data into that string.

IMO there's nothing wrong with asking the user to import `toBase64FromMedia` from the correct package (`web/` or `node/`), especially given it's only used for (technically) one query type. It simplifies the code massively tho.

packages/core/src/index.ts
```diff
@@ -137,106 +129,60 @@ const cleanHost = (host: string, protocol: 'rest' | 'grpc') => {
...
-    apiKey: isApiKey(params.auth) ? mapApiKey(params.auth) : undefined,
-    authClientSecret: isApiKey(params.auth) ? undefined : params.auth,
-    agent,
-    timeout: params.timeout,
-    skipInitChecks: params.skipInitChecks,
-  });
+  const agent = context.agentMaker?.(httpSecure);
+
+  const { connection, dbVersionProvider, dbVersionSupport } = await ConnectionGRPC.use(
+    context.transportsMaker,
```
bevzzz, line 162: tl;dr: I find the fact that `ConnectionGRPC` satisfies `ConnectionREST` _so confusing_.

---

Initially I wanted to ask this:

> question: if `agentMaker` "Creates the HTTP(S) keep-alive agent for the REST/GraphQL connection", then why is it a part of the top-level Context and not some internal component of ConnectionREST?

Before posting that, I decided to look around and answer this for myself. But I couldn't find ConnectionREST mentioned anywhere here, and instead I saw `connection` returned by `ConnectionGRPC.use()` passed to `alias(connection)` which expects ConnectionREST (!).

I was rather surprised to learn that ConnectionGRPC _extended_ ConnectionGQL even though they use completely different transports, and that ConnectionGQL was a thing separate from ConnectionREST even though both use _the same transport / connection_.

---

suggestion (wrt `agentMaker`): I see from it's description that `web/` and `node/` inject different agent makers. Perhaps each could instead provide a complete `ConnectionREST` object and make the injection "privately"?

packages/core/src/index.ts
```diff
@@ -137,106 +129,60 @@ const cleanHost = (host: string, protocol: 'rest' | 'grpc') => {
...
+    {
+      host: `${scheme}://${httpHost}:${httpPort}${httpPath || ''}`,
+      scheme: scheme,
+      headers: params.headers,
+      grpcAddress: `${isGrpcWeb ? (grpcSecure ? 'https://' : 'http://') : ''}${grpcHost}:${grpcPort}${
```
bevzzz, line 167: question: why are there both `httpSecure` and `grpcSecure` parameters? Is there a real use-case for having one connection be secured and the other one not?

---

```suggestion
      grpcAddress: `${isGrpcWeb ? (grpcSecure ? 'https' : 'http') : ''}://${grpcHost}:${grpcPort}${
```
nit: simplify the expression

packages/core/src/serialize/index.ts
```diff
@@ -2070,7 +2082,7 @@ export class Serialize {
   };
 
   public static tenants<T, M>(tenants: T[], mapper: (tenant: T) => M): M[][] {
-    const mapped = [];
+    const mapped: M[][] = [];
     const batches = Math.ceil(tenants.length / 100);
     for (let i = 0; i < batches; i++) {
       const batch = tenants.slice(i * 100, (i + 1) * 100);
```
bevzzz, line 2203: question: why does this method split the tenants array into batches or 100?

packages/test/node/collections/integration.test.ts
```diff
@@ -246,6 +249,7 @@ describe('Testing of the collections.create method', () => {
 
   it('should be able to create a simple collection with a generic using a schema var with type', async () => {
     const collectionName = 'TestCollectionSimpleGenericVarType';
+    afters.push(() => contextionary.collections.delete(collectionName));
```
bevzzz, line 252: suggestion: would it make sense to add a test util for creating collections that would "schedule" the deletion automatically?

It's exactly as what you have with `afters` here, but packaged so that it's easier to reuse across tests and a tad more succinct. Just something I've thought about doing every time I had to create a collection in a test :)

---

```ts
// testutil.ts
export class Collectioner {
  private client WeaviateClient;
  private collections: []string;

  public create(c: CollectionConfig): Promise<CollectionHandle> {
    this.collections.push(c.name);
    return this.client.collections.create(c);
  }
  
  public cleanup(): Promise<void> {
      return Promise.all(this.collections.map(c => this.client.collections.delete(c)));
  }
}
```

Then use it in a test:

```ts
describe('stuff', async () => {
  const client = new WeaviateClient(...);
  const collections = new Collectioner(client);
  afterAll(() => collections.cleanup());
  
  ...
})
```

packages/web/src/index.ts
```diff
@@ -0,0 +1,104 @@
...
+const webify = (
+  context: Context<string | Blob>,
+  params: ClientParams
+): Promise<IWeaviateClient<string | Blob>> => {
+  params.connectionParams.grpc = {
+    host: params.connectionParams.http.host,
+    port: params.connectionParams.http.port,
+    secure: params.connectionParams.http.secure,
+    path: '/v1/grpc-web',
+  };
```
bevzzz, line 43: question: could host, port, and secure be mapped in `helpers`? Seems like something that'd be common to both node and web, as opposed to `path: /v1/grpc-web` which is clearly specific to `web/` package.

packages/core/src/collection/index.ts
bevzzz: Having went through the whole PR I want to emphasize `much easier` :)

parkerduckworth:
Let's go @tsmith023 !! The time for a TS web client has finally come ✨

packages/core/src/connection/grpc.ts
tsmith023: have simplified this area of the codebase, lmk wyt

packages/core/src/collection/index.ts
tsmith023: ye :( this generic change is necessary to allow for drilling the different media types from node/web but certainly many changes have been lumped into this one mega PR

packages/core/src/connection/grpc.ts
tsmith023: this is a good point and raises the fact that we should allow users to specify the max message size if they know it rather than requiring this init check. Then they'd be able to set it once on the server + client and dodge init checks to speed up queries, which I've added now

packages/core/src/connection/grpc.ts
tsmith023: good catch, have fixed

packages/core/src/index.ts
tsmith023: yep have fixed

packages/core/src/serialize/index.ts
tsmith023: iirc, to avoid overloading the db but any kind of batching like this would be better placed on the server rather than in the client tbh

packages/test/node/collections/integration.test.ts
tsmith023: I like the idea of doing this in a dedicated cleanup PR. The python client has something similar so would make sense to introduce that pattern here too

packages/web/src/index.ts
tsmith023: This is mapping the provided `http` host, port, secure onto the `grpc` ones so is specific to `grpc-web`

packages/core/src/index.ts
tsmith023: yeh it's very confusing and is one of the worst parts of the TS client

the reason for it was to maintain bc/fc between the v3 and v2 clients when shipping both at the same time and building the simpler v3 APIs off the old v2 APIs, e.g. `collection.tenants.create` uses `TenantsCreator` under-the–hood, a decision made to speed up development in the pre-agentic era!

tsmith023: if we spend an effort removing all the old v2 logic then we could do away with `ConnectionREST` and `ConnectionGQL` thereby simplifying the entire codebase

packages/core/src/index.ts
tsmith023: agreed in principle, the thing is that this is a feature of the API currently so needs to be maintained moving forward if we want to achieve backwards compatibility

bevzzz: I understand (and, largely, respect) the b/c argument. But will the new version (after this PR) be backwards compatible in the sense that the user doesn't need to adapt their code after they'd upgraded to the latest version? Surely not.

There're now 2 sub-packages `web/` and `node/`, each with its own set of `connectToLocal`/`connectToWcd`, so a user upgrading to the latest version already has to go and change:

```diff
- import { connectToLocal } from 'weaviate-client';
+ import { connectToLocal } from 'weaviate-client/node';
```

If so, I don't see it as an issue to require the following change too, if it means we can avoid threading a third-order-importance function through the entire package.

```diff
+ import { toBase64FromMedia } from 'weaviate-client/node';

let Buffer b = loadImage();
- nearMedia({ image: b });
+ nearMedia({ image: toBase64FromMedia(b) });
```

WDYT? 

I might've accidentally glossed over an important detail, please correct me if that's so! The above is my understanding.

tsmith023: it's BC for `weaviate-client`, which will continue to be released and will be equivalent to `@weaviate/node` in the short-term

my thinking now is that we actually leave `weaviate-client` on a dedicated branch and have `main` refer to the `@weaviate/node` and `@weaviate/web` pkgs

the problem with this will be when adding features to `@weaviate/node` we'd need to backport them to `weaviate-client` in a non-trivial way in the short-term

we did this with the v3 & v2 clients while people migrated over but this time `weaviate-client` is used much more than `weaviate-ts-client` three years ago

packages/core/src/connection/grpc.ts
bevzzz: > to speed up queries
Probably won't speed the queries up _that much_ unless a new client is created per query? Might be the case for some serverless, Lambda-style deployments. 🤷 

Good addition overall.

PR #315, opened 2025-07-10
feat: Add support for RQ

bevzzz:
RQ can be configured so:

```ts
weaviate.configure.vectorIndex.quantizer.rq({ bits: 6, rescoreLimit: 20 });
```

When updating, it is only possible to update `rescoreLimit`.

### DIFF
--- a/src/collections/config/types/vectorIndex.ts
+++ b/src/collections/config/types/vectorIndex.ts
@@ -9,7 +9,7 @@ export type VectorIndexConfigHNSW = {
   filterStrategy: VectorIndexFilterStrategy;
   flatSearchCutoff: number;
   maxConnections: number;
-  quantizer: PQConfig | BQConfig | SQConfig | RQConfig | undefined;
+  quantizer: QuantizerConfig | undefined;
   skip: boolean;
   vectorCacheMaxObjects: number;
   type: 'hnsw';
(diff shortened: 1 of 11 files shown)


src/collections/config/types/vectorIndex.ts
```diff
@@ -9,7 +9,7 @@ export type VectorIndexConfigHNSW = {
   filterStrategy: VectorIndexFilterStrategy;
   flatSearchCutoff: number;
   maxConnections: number;
-  quantizer: PQConfig | BQConfig | SQConfig | undefined;
+  quantizer: PQConfig | BQConfig | SQConfig | RQConfig | undefined;
```
tsmith023, line 12: Looking at this with fresh eyes, should we use `QuantizerConfig` here to reduce the duplication? So: `quantizer: QuantizerConfig | undefined`?

src/collections/config/utils.ts
```diff
@@ -468,11 +479,13 @@ class ConfigMapping {
       throw new WeaviateDeserializationError(
         'Vector index vector cache max objects was not returned by Weaviate'
       );
-    let quantizer: PQConfig | BQConfig | SQConfig | undefined;
+    let quantizer: PQConfig | BQConfig | SQConfig | RQConfig | undefined;
```
tsmith023, line 482: Same here for `QuantizerConfig`

src/collections/config/types/vectorIndex.ts
bevzzz: good idea 👍

tsmith023 (approved):

PR #316, opened 2025-07-11
feat: Collection aliasing

bevzzz:
> ➡️ Merge after #315. There are some CI configuration updates and test fixes.

This PR adds API for managing collection aliases and extends the set of permissions with AliasPermission.

Manage aliases

```ts
client.alias.create({ collection: "Pizza", alias: "PizzaPie" });
client.alias.listAll();
client.alias.listAll("Pizza"); // list aliases defined for Pizza collection
client.alias.get("PizzaPie");
client.alias.update({ alias: "PizzaPie", newTargetCollectionName: "Langos" });
client.alias.delete("PizzaPie");
```

Manage alias permissions

```ts
weaviate.permissions.aliases({
    alias: ["PizzaPie", "PizzaWhy"],
    create: true,
    read: true,
    update: true,
    delete: true,
})
```

### DIFF
--- a/src/alias/index.ts
+++ b/src/alias/index.ts
@@ -1,6 +1,6 @@
 import { ConnectionREST } from '../index.js';
 import { WeaviateAlias, WeaviateAliasResponse } from '../openapi/types.js';
-import { Alias, AliasListAllOpts, CreateAliasArgs, UpdateAliasArgs } from './types.js';
+import { Alias, AliasListAllOptions, CreateAliasArgs, UpdateAliasArgs } from './types.js';
 
 export interface Aliases {
   /**
@@ -21,7 +21,7 @@ export interface Aliases {
    * @param {string | undefined} [opts.collection] Get all aliases defined for this collection.
    * @returns {Promise<Alias[] | undefined>} An array of aliases.
    */
-  listAll: (opts?: AliasListAllOpts) => Promise<Alias[] | undefined>;
+  listAll: (opts?: AliasListAllOptions) => Promise<Alias[] | undefined>;
 
   /**
    * Get information about an alias.
@@ -56,7 +56,7 @@ const alias = (connection: ConnectionREST): Aliases => {
   return {
     create: (args: CreateAliasArgs) =>
       connection.postReturn<WeaviateAlias, void>(`/aliases/`, { ...args, class: args.collection }),
-    listAll: (opts?: AliasListAllOpts) =>
+    listAll: (opts?: AliasListAllOptions) =>
       connection
         .get<WeaviateAliasResponse>(
           `/aliases${opts?.collection !== undefined ? '/?class=' + opts.collection : ''}`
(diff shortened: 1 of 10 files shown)


src/alias/index.ts
```diff
@@ -0,0 +1,77 @@
...
+   * Create alias for a collection.
+   *
+   * The collection must exist prior to aliasing it.
+   * One alias cannot be created for multiple collections simultaneously.
+   *
+   * @param {string} [opt.collection] Original collection name.
+   * @param {string} [opt.alias] Alias for collection.
+   * @returns {Promise<void>} Awaitable promise.
+   * */
+  create: (opt: CreateAliasInput) => Promise<void>;
```
tsmith023, line 16: Imo, since `opt` here is required (doesn't allow undefined `opt?`) and `collection` + `alias` inside the type are also not able to `undefined`, it should be named `args: CreateAliasArgs` to align with other areas of the codebase, e.g. `Backup.cancel`, wdyt?

In addition, required params in TSdoc should not have the `[]` wrapping. I think there are issues elsewhere in the codebase outside of this PR with the consistency of this but we should make sure new code follows it regardless 😁

src/alias/index.ts
```diff
@@ -0,0 +1,77 @@
...
+   * Replace target collection the alias points to.
+   *
+   * To change the alias that points to the collection,
+   * delete the alias and create a new one.
+   *
+   * @param {string} [opt.alias] Alias to update.
+   * @param {string} [opt.collection] New collection the alias should point to.
+   * @return {Promise<void>} Awaitable promise.
+   */
+  update: (opt: UpdateAliasInput) => Promise<void>;
```
tsmith023, line 44: Same as for `create` above

src/alias/index.ts
```diff
@@ -0,0 +1,77 @@
...
+   * @param {string} [opt.collection] Original collection name.
+   * @param {string} [opt.alias] Alias for collection.
+   * @returns {Promise<void>} Awaitable promise.
+   * */
+  create: (opt: CreateAliasInput) => Promise<void>;
+
+  /**
+   * List all aliases defined in the schema.
+   *
+   * @param {string | undefined} collection Get all aliases defined for this collection.
```
tsmith023, line 21: To align with TSdoc convention, this should be `[collection]`

src/alias/index.ts
```diff
@@ -0,0 +1,77 @@
...
+   * @returns {Promise<Alias[] | undefined>} An array of aliases.
+   */
+  listAll: (collection?: string) => Promise<Alias[] | undefined>;
```
tsmith023, line 24: Since `collection?` is optional here, what do you think about introducing a type, say `AliasListAllOpts` in case we ever need to extend this API to include other query options when listing aliases?

tsmith023 (changes requested):

tsmith023 (approved):

PR #317, opened 2025-07-11
feat: add support for multi2multivec-jinaai vectorizer

bevzzz:
> ➡️ Merge after https://github.com/weaviate/typescript-client/pull/315. There are some CI configuration updates and test fixes.

Weaviate 1.32 is going to feature [`multi2multivec-jinaai` module](https://github.com/weaviate/weaviate/pull/8609). With this PR it will be possible to configure it in a TS project.

Use it like this: 

```ts
configure.vectorizer.multi2MultivecJinaAI({
    name: 'multi-jina',
    imageFields: ['field1', 'field2'],
    textFields: ['field3', 'field4'],
})
```


antas-marcin (approved):

tsmith023:
Given the introduction of the `configure.multiVectors` factory here https://github.com/weaviate/typescript-client/pull/264, I think this PR should be modified to match the new API. wdyt? @bevzzz

bevzzz:
> Given the introduction of the configure.multiVectors factory in https://github.com/weaviate/typescript-client/pull/264

@tsmith023 since those changes haven't been released yet, do you think it'll make sense to rename the method to `text2MultivecJinaAI`? Because the module is called `text2multivec-jinaai` and the convention is to name the static factory after the module.

I think it's more consistent this way

tsmith023:
> multiVectors.text2VecJinaAI to multiVectors.text2MultivecJinaAI

@bevzzz, the idea of dropping the `MultiVec` here is because it is already used in the `multiVectors` factory name so is redundant information. I would agree that it should be `text2MultiVecJinAI` if the method was under `vectors`, however

bevzzz:
> the idea of dropping the MultiVec here is because it is already used in the multiVectors factory name

I renamed the static factory for the new module to `multi2VecJinaAI` to be consistent with the other example

tsmith023 (approved):

PR #321, opened 2025-07-21
Fix breaking changes from v3.7.0 release

bevzzz:
We've made a number of breaking changes in the `v3.7.0` release, which we're going to deprecate in favor of `v3.8.0`. 
This PR brings back removed exports and deprecates `vectorizeCollectionName` parameter for all vectorizer modules where it has previously been accepted. See comment on `legacyVectors` for more details.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -129,7 +129,7 @@ jobs:
           registry-url: 'https://registry.npmjs.org'
       - run: npm ci
       - run: npm run build
-      - run: npm publish --tag alpha
+      - run: npm publish
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_AUTOMATION_TOKEN }}
       - run: npm run docs


src/collections/configure/index.ts
```diff
@@ -311,5 +311,6 @@ export {
   tokenization,
   vectorDistances,
   configureVectorIndex as vectorIndex,
+  legacyVectors as vectorizer,
```
bevzzz, line 314: This fixes another [breaking change](https://github.com/weaviate/typescript-client/pull/264/files?file-filters%5B%5D=.ts&show-viewed-files=true#diff-a3357c4e265a0992e8b0b8168ee30f635918ffdcfa03edb3db9c830b955d00ecL312-R314): in addition to maintaining `configure.vectorizer` we have to maintain `vectorizer` export.

src/collections/config/types/vectorizer.ts
```diff
@@ -393,8 +393,6 @@ export type Text2MultiVecJinaAIConfig = {
   dimensions?: number;
   /** The model to use. */
   model?: string;
-  /** Whether to vectorize the collection name. */
-  vectorizeCollectionName?: boolean;
```
bevzzz, line 397: This type has been first introduced in 3.7.0. As we're deprecating that version entirely, might as well delete the unwanted config parameter.

src/collections/configure/generative.ts
```diff
@@ -262,7 +269,14 @@ export default {
   ): ModuleConfig<'generative-google', GenerativeGoogleConfig | undefined> => {
     return {
       name: 'generative-google',
-      config,
+      // Do not populate config key if config === undefined.
+      config: config
+        ? {
+            ...config,
+            // Only create modelId key if either config.model or config.modelId are defined.
+            ...(config?.modelId || config?.model ? { modelId: config?.model ?? config?.model } : undefined),
```
bevzzz, line 277: This syntax is fairly dense. My intention was to copy the model name into the `modelId` variable (for b/c) but preserve the previous behavior for `config === undefined`

src/collections/config/types/vectorizer.ts
malgamves: Is it unwanted? Are there any cases someone might want this ?

src/collections/configure/index.ts
malgamves: Good catch! Very true

src/collections/config/types/vectorizer.ts
bevzzz: We're doing this as part of an effort to improve the UX around vectorization. Long story short setting this to `true` can get Weaviate to produce useless vectors for collections that actually have no embeddable properties.

Instead of returning an error on write, we would just generate an embedding of the collection name every time. 

It is also meaningless to some of the modules, like `text2multivec-jinaai` in the example above, but it made its way to the API cuz it seemed plausible :)

mpartipilo (approved):

PR #322, opened 2025-07-21
fix: export vectorizer from configure/index.js

bevzzz:
In https://github.com/weaviate/typescript-client/pull/264 we've unintentionally removed `vectorizer` from the list of exports in `collections/configure/index.js`, which is a breaking change.

This PR brings it back and marks the export as deprecated.


bevzzz:
Done in #321

PR #330, opened 2025-08-13
feat: support list backups

bevzzz:
Weaviate 1.32 supports listing all backups for a backend. This functionality's already been implemented in the client, but we've refrained from exporting before.

Now `BackupGetter` joins the public API for the TS client.

### DIFF
--- a/src/collections/backup/integration.test.ts
+++ b/src/collections/backup/integration.test.ts
@@ -131,9 +131,7 @@ describe('Integration testing of backups', () => {
               includeCollections: ['TestListBackups'],
               waitForCompletion: true,
             })
-            .then((res) => {
-              return res.id;
-            })
+            .then((res) => res.id)
         );
       }
 


tsmith023:
This code looks to be relevant only to the old v2 client, which we haven't released a new version of for a year. For the v3 client, we'd need to add a `list` method to the interface [here](https://github.com/weaviate/typescript-client/blob/main/src/collections/backup/client.ts#L208)

src/collections/backup/integration.test.ts
```diff
@@ -88,6 +89,23 @@ describe('Integration testing of backups', () => {
...
+        backend: res.backend as Backend,
+      });
+      if (status === 'SUCCESS') {
+        wait = false;
+      }
+      if (status === 'FAILED') {
+        throw new Error(`Backup creation failed: ${error}`);
+      }
+      await new Promise((resolve) => setTimeout(resolve, 1000));
+    }
```
bevzzz, line 107: Added this wait to make sure the in-progress backup does not interfere with any tests that might follow.

PR #358, opened 2025-10-22
feat: add return fields for list backups and support ordering

gkampitakis:
Adds supports for this change https://github.com/weaviate/weaviate/pull/9502

### DIFF
--- a/src/collections/backup/client.ts
+++ b/src/collections/backup/client.ts
@@ -28,6 +28,7 @@ import {
   BackupReturn,
   BackupStatusArgs,
   BackupStatusReturn,
+  ListBackupOptions,
 } from './types.js';
 
 export const backup = (connection: Connection): Backup => {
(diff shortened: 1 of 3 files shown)


src/collections/backup/client.ts
```diff
@@ -261,7 +265,8 @@ export interface Backup {
   /** List existing backups (completed and in-progress) created in a given backend.
    *
    * @param {Backend} backend Backend whence to list backups.
+   * @param {sortAscending} [boolean] Sort list of backups in ascending order based on creation time. Default is descending order.
    * @returns {Promise<BackupReturn[]>} The response from Weaviate.
    * */
-  list(backend: Backend): Promise<BackupReturn[]>;
+  list(backend: Backend, sortAscending?: boolean): Promise<BackupReturn[]>;
```
bevzzz:
```suggestion
  list(backend: Backend, opts?: ListBackupOptions): Promise<BackupReturn[]>;
```

where ListBackupOptions is defined as follows:

```ts
type ListBackupOptions = { startedAtAsc?: boolean };
```


Two things here:

1. I suggest mentioning the field backups are sorted by in the variable name, i.e. "started at", similarly to what's been done in Go client
2. A pattern for any optional parameters we use across the repo is to declare an -Options type and make it the last optional argument to the function.

PR #380, opened 2025-12-29
Object TTL

bevzzz:
This PR adds support for Object TTL configuration in collections.


src/collections/configure/index.ts
```diff
@@ -117,6 +119,26 @@ const configure = {
           : undefined,
     };
   },
+  objectTTL: (options?: {
+    enabled?: boolean;
```
dirkkul: I also would not add enabled as an option here. If you set TTL for creation it is enabled

bevzzz:
`collections/journey.test.ts` integration test is failing due to current maintenance and testing in WCD dev clusters

PR #385, opened 2026-01-16
Inject client version header in API requests

mpartipilo:
Inject the client version into requests by defining `WEAVIATE_CLIENT_VERSION` during the build process. Update the HTTP and gRPC clients to include this version in the request headers. Add tests to verify the correct inclusion of the client version header in requests.


scripts/gen-cersion.mjs
```diff
@@ -0,0 +1,7 @@
+import fs from 'fs';
+import pkg from '../package.json' with { type: 'json' };
+
+fs.writeFileSync(
+  'src/version.ts',
+  `export const WEAVIATE_CLIENT_VERSION = '${pkg.version}';\n`
+);
```
mpartipilo: Did you probably mean to name the file `gen-version.mjs`?

PR #386, opened 2026-01-19
Feat: introduce batch data ingestion

tsmith023:


src/collections/data/batch.ts
```diff
@@ -0,0 +1,462 @@
...
+        batchingErr = err;
+      });
+      const check = (err: Error | null) => {
+        if (err) {
+          throw err;
+        }
+        return batcher;
+      };
+      return {
+        addObject: (obj) => check(batchingErr).addObject(obj),
```
bevzzz: suggestion: iiuc the expectation that `batcher.start()` will be resolved before the returned `Batch` object is used is _implicit_ -- `check(batchingErr)` will not throw if the promise hadn't been resolved previously.

since `stream` returns a promise already, why don't we connect the events explicitly:

```ts
return batcher.start(connection)
  .then(conn -> new Batcher<any>(...))
  .then(batcher -> {
    addObject: (obj) => batcher.addObject(obj),
    ...,
  });
```

The error from `start` can be handled by the user via `.catch()` callback or the usual try-catch.

src/collections/data/batch.ts
```diff
@@ -0,0 +1,462 @@
...
+  public uuids: Record<number, string> = {};
+  public beacons: Record<number, string> = {};
+
+  constructor(args: BatcherArgs) {
+    this.consistencyLevel = args.consistencyLevel;
+    this.queue = new Queue();
+    this.isGcpOnWcd = args.isGcpOnWcd;
+  }
+
+  private healthy() {
```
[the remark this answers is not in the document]
bevzzz: Alternatively, `

src/collections/data/batch.ts
```diff
@@ -0,0 +1,462 @@
...
+  private consistencyLevel?: ConsistencyLevel;
+  private queue: Queue<BatchObject<T> | BatchReference>;
+
+  private inflightObjs: Set<string> = new Set();
+  private inflightRefs: Set<string> = new Set();
+  private batchSize = 1000;
+  private objsCache: Record<string, Internal<BatchObject<T>>> = {};
+  private refsCache: Record<string, Internal<BatchReference>> = {};
+  private pendingObjs: BatchObjectGRPC[] = [];
+  private pendingRefs: BatchReferenceGRPC[] = [];
```
bevzzz: nit: `Objs` is not much shorted than `Objects` but reads better (in the long run).

I can see the value of having `refs` and `objs` be aligned (4 chars each), so feel free to ignore.

src/collections/data/batch.ts
```diff
@@ -0,0 +1,462 @@
...
+
+        yield BatchStreamRequest.create({ stop: {} });
+        return;
+      }
+
+      if (isBatchObject<T>(entry)) {
+        const { grpc } = Serialize.batchObject<T>(entry.collection, entry, false, entry.tenant);
+        this.objsCache[grpc.uuid] = { entry, index: objIndex };
+
+        const objSize = BatchObjectGRPC.encode(grpc).finish().length + perObjectOverhead;
```
bevzzz: nit: `perObjectOverhead` is defined far from the place where it's used

suggestion: define it right before the current block or as package-level constant `PER_OBJECT_OVERHEAD`

src/collections/data/batch.ts
```diff
@@ -0,0 +1,462 @@
...
+        if (totalSize + objSize >= grpcMaxMessageSize || req.data!.objects!.values.length >= this.batchSize) {
+          while (this.inflightObjs.size > this.batchSize) {
```
bevzzz: question: should this be `>=` like above in `addObject` and `addReference`?

src/collections/data/batch.ts
```diff
@@ -0,0 +1,462 @@
...
+export default function (connection: Connection, dbVersionSupport: DbVersionSupport): Batch {
+  return {
+    stream: async (consistencyLevel) => {
+      const { supports, message } = await dbVersionSupport.supportsServerSideBatching();
+      if (!supports) {
+        throw new Error(message);
+      }
+      const batcher = new Batcher<any>({ consistencyLevel, isGcpOnWcd: connection.isGcpOnWcd() });
+      let batchingErr: Error | null = null;
+      const batching = batcher.start(connection).catch((err) => {
```
bevzzz: note: I was initially confused to see `stop()` return `batching`, because it looked like the promise would resolve once it's been _started_ and not when the entire "batching context" was closed.

```suggestion
      const context = batcher.start(connection).catch((err) => {
```

src/collections/data/index.ts
```diff
@@ -83,6 +85,7 @@ export interface Data<T> {
     opts?: DeleteManyOptions<V>
   ) => Promise<DeleteManyReturn<V>>;
   exists: (id: string) => Promise<boolean>;
+  ingest: (objs: Iterable<DataObject<T> | NonReferenceInputs<T>>) => Promise<BatchObjectsReturn<T>>;
```
bevzzz: question: `ingest` looks like a convenience wrapper around Batcher -- why doesn't it support references?

bevzzz (commented):
Some of the suggestions are influenced by the design decisions I made whilst working on the Java implementation, so they might not translate well -- your call.

src/collections/data/batch.ts
```diff
@@ -0,0 +1,462 @@
...
+  private refsCache: Record<string, Internal<BatchReference>> = {};
+  private pendingObjs: BatchObjectGRPC[] = [];
+  private pendingRefs: BatchReferenceGRPC[] = [];
+  private isStarted = false;
+  private isShuttingDown = false;
+  private isOom = false;
+  private isStopped = false;
+  private isGcpOnWcd = false;
+  private isRenewingStream = false;
+  private oomWaitTime = 300;
```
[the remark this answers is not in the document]
tsmith023: `oomWaitTime` will be undefined until it is set by the message. I set a default here to avoid the typing issues associated with `oomWaitTime` having type `number | undefined`

src/collections/data/batch.ts
tsmith023: yes, good catch!

src/collections/data/batch.ts
```diff
@@ -0,0 +1,462 @@
...
+  private objsCache: Record<string, Internal<BatchObject<T>>> = {};
+  private refsCache: Record<string, Internal<BatchReference>> = {};
+  private pendingObjs: BatchObjectGRPC[] = [];
+  private pendingRefs: BatchReferenceGRPC[] = [];
+  private isStarted = false;
+  private isShuttingDown = false;
+  private isOom = false;
+  private isStopped = false;
+  private isGcpOnWcd = false;
+  private isRenewingStream = false;
```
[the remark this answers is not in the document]
tsmith023: These booleans aren't mutually exclusive, some can be `true` at the same time, e.g. `isStopped` can be `true` while `isShuttingDown` is true. The ones that can be unified are `isShuttingDown` and `isOom` but I don't see the value in introducing `state` just for them, wdyt?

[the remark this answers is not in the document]
bevzzz: yep, certainly not worth it if it's only 2/5 states

src/collections/data/batch.ts
```diff
@@ -0,0 +1,471 @@
...
+        yield BatchStreamRequest.create({ stop: {} });
+        return;
+      }
+
+      if (isBatchObject<T>(entry)) {
+        const { grpc } = Serialize.batchObject<T>(entry.collection, entry, false, entry.tenant);
+        this.objsCache[grpc.uuid] = { entry, index: objIndex };
+
+        const objSize = BatchObjectGRPC.encode(grpc).finish().length + PER_OBJECT_OVERHEAD;
+        if (objSize + PER_OBJECT_OVERHEAD + emptyReqSize >= grpcMaxMessageSize) {
```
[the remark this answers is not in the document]
tsmith023: ty!

src/collections/data/batch.ts
```diff
@@ -0,0 +1,471 @@
...
+        objErrors: () => batcher.objErrors,
+        refErrors: () => batcher.refErrors,
+      };
+    },
+  };
+}
+
+type BatcherArgs = {
+  consistencyLevel?: ConsistencyLevel;
+  isGcpOnWcd: boolean;
```
bevzzz: nit: should this be `isWcdOnGcp` since it's _running on GCP_?

PR #387, opened 2026-01-19
Reorganise repo to use `vitest` instead of `jest`

tsmith023:
This PR refactors and reorganises the repository to better isolate tests from the source of the package while also migrating from `jest` to `vitest`. In doing so, it also bumps dependencies, e.g. of `typescript`, and removes unused/outdated packages. Any issues in the src found through the test refactoring are also fixed.

### DIFF
--- a/src/collections/configure/index.ts
+++ b/src/collections/configure/index.ts
@@ -20,7 +20,7 @@ import { multiVectors, vectorizer, vectors } from './vectorizer.js';
 
 import { parseWithDefault } from './parsing.js';
 
-export const dataType = {
+const dataType = {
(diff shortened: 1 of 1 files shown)


bevzzz:
> reorganises the repository to better isolate tests from the source of the package 

opinion (non-blocking): I don't know that this is necessarily an improvement for _unit tests_.
Those, I feel, should serve as "documentation" for the code they're testing and many modern languages encourage keeping them close (Go with thing.go and thing_test.go) or even in the same source file (Rust / Zig)

Integration tests are a different beast, so no objections there.

PR #401, opened 2026-02-20
Add support for cancelling backup restore operations

tsmith023:


src/collections/backup/types.ts
```diff
@@ -85,6 +78,8 @@ export type BackupCancelArgs = {
   backupId: string;
   /** The backend to use for the backup. */
   backend: Backend;
+  /** The type of operation to cancel (backup creation or restoration). Defaults to 'backup'. */
+  type?: 'backup' | 'restore';
```
bevzzz: nit: in the codebase "backup" is mostly used as a noun (backed-up files) and "create" as the verb for creating it. I'd also argue that _operation_ is more descriptive than "type"

```suggestion
  operation?: 'create' | 'restore';
```

tsmith023: Agreed!

PR #416, opened 2026-04-14
Add capability for requesting profiling of queries

tsmith023:


src/collections/serialize/index.ts
```diff
@@ -565,7 +565,7 @@ class Search {
         score: true,
         explainScore: true,
         isConsistent: true,
-        queryProfile: false,
+        queryProfile: true,
```
g-despot, line 595: This should probably be `false`, profiling adds a lot of overhead and should always be opt-in.

test/collections/query/integration.test.ts
```diff
@@ -232,6 +246,41 @@ describe('Testing of the collection.query methods with a simple collection', ()
     expect(ret.objects[0].properties.testProp).toEqual('carrot');
     expect(ret.objects[0].uuid).toEqual(id);
   });
+
+  requireAtLeast(1, 36, 0).it('should return query profiling metadata with filtered fetch', async () => {
```
g-despot, line 250: Available from 1.36.9, but I see the tests are passing so I assume it uses always the latest patch version.

g-despot (changes requested):

test/collections/query/integration.test.ts
tsmith023: Ye this is just in a test so the exact patch doesn't matter. In the client, if they're using an old server then the return simply won't contain any profiling data so it's BC

src/collections/types/query.ts
```diff
@@ -21,9 +35,9 @@ export type Metadata = {
   isConsistent: boolean;
 };
 
-export type MetadataKeys = (keyof Metadata)[];
+export type MetadataKeys = (keyof Metadata | 'queryProfile')[];
 
-export type QueryMetadata = 'all' | MetadataKeys | undefined;
+export type QueryMetadata = 'all' | 'all+queryProfile' | MetadataKeys | undefined;
```
bevzzz: issue: The `all+queryProfile` looks awkward, esp. considering a possibility of another "expensive" parameter being added in the future. In which case we'll end up with a combination of `all`, `all+x`, `all+y`, `all+x+y`, and so on.

suggestion: `MetadataKey` already allows passing _several_ options at once. I think all it takes to accommodate an "optional `all`" is this:

```ts
export type MetadataKeys = (keyof Metadata | 'all' | 'queryProfile')[];
```

The user can combine these as they like:

```ts
{ returnMetadata: 'all' }
{ returnMetadata: ['all', 'queryProfile'] }
```

bevzzz (changes requested):

src/collections/types/query.ts
tsmith023: Ye I considered this but it does allow:
```typescript
{ returnMetadata: ['all'] }
```
which I didn't like. But perhaps it's a worthwhile tradeoff to avoid `all+queryProfile`, wdyt?

bevzzz: > Ye I considered this but it does allow:
> ```ts
> { returnMetadata: ['all'] }
> ```

Why not? Here's a [playground](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAshwEMAmDFQLxQN5QK64EskAuKAIwHsKAbCBAOwG4oBjAJzuAiQEFhTKNOvSgBfRgCgJoSLHjJUCANIQQAZwxQAFAGtVFAGZzEKNFAA+UAOQJq1KxesBHXBDYgACm0MFaVgJQA2gC6ktLg0ABK8Lhs9HAmipo2dg6WCQqIKuqOuPRIEAYE9NxhLBT0asBQLm4gmlpsALYA-KTRwLHx8qYI-hgAfNiiUrXuWv6SYyBaKfaTEtNagXNWwQtLK-hEVgA01uyc3HxrG67jK7b2+1bTXj5+64xAA) with an example.

tsmith023: Just that we'd allow two ways of doing the same thing `returnMetadata: 'all'` and `returnMetadata: ['all']`, which should be avoided as a rule imo. In this case, it helps to avoid `all+queryProfile` so it's worth it so will change

bevzzz: > Ye I considered this but it **does** allow:

Oh-oh, sorry, I read "does not allow" the first time around! It's why I attached the playground example, to show that it's fine with the type system. 

> two ways of doing the same thing `returnMetadata: 'all'` and `returnMetadata: ['all']`, which should be avoided as a rule imo

though fair point, I also feel `all+queryProfile` is just much more awkward

bevzzz (approved):

PR #418, opened 2026-04-15
1.37/support drop vector index

tsmith023:


src/collections/config/index.ts
```diff
@@ -173,6 +175,17 @@ export interface Config<T> {
    * @returns {Promise<void>} A promise that resolves when the index has been dropped.
    */
   dropInvertedIndex: (propertyName: string, indexName: InvertedIndexName) => Promise<void>;
+  /**
+   * Drop the vector index associated with a named vector of this collection.
+   *
+   * This is a destructive operation. The index will need to be regenerated if you wish to use it again.
+   *
+   * @param {keyof V & string} [vectorName] The name of the vector to drop the index from. If not provided, the default vector index will be dropped.
```
bevzzz, line 183:
```suggestion
   * @param {keyof V & string} [vectorName] The name of the vector to drop the index from. If not provided, the "default" vector index will be dropped.
```

Maybe a nit, but I think it's good to be specific about the fact that Weaviate doesn't have the notion of a default vector index, and it's the client that will fallback to `"default"` input. E.g. if the user's only index is called "my-only-index" they should expect to see an error on `dropVectorIndex(undefined)`.

I would argue this parameter should be required tho, similarly to how when you're trying to delete a GH repo you need to manually enter the repo's name to confirm.

test/collections/config/integration.test.ts
```diff
@@ -1085,6 +1085,76 @@ describe('Testing of the collection.config namespace', () => {
...
+      const collection = await client.collections.create({
+        name: collectionName,
+        vectorizers: weaviate.configure.vectors.selfProvided({
+          name: 'existing',
+          vectorIndexConfig: weaviate.configure.vectorIndex.hnsw(),
+        }),
+      });
+
+      await expect(collection.config.dropVectorIndex('nonexistent')).rejects.toThrow(
+        'The request to Weaviate failed with status code: 422 and message: {"error":[{"message":"not found: vector index \\"nonexistent\\" not found in class \\"TestDropNonExistentNamedVectorIndex\\""}]}'
```
bevzzz, line 1118: Does the client generally throws errors on the 200+ HTTP codes? If yes, then it seems like this test is redundant, it essentially tests the server's response to a non-existen input. If no, then let's not assert the contents of the message -- someone may re-phrase that in v1.38 or sth and fail the client pipeline (used to happen a lot in the Java client test suite)

PR #419, opened 2026-04-15
Re-export `WeaviateClass` from where it is used in the public API

### LINKED ISSUE #404: Types: Please export public types
While starting work on a new project that is using weaviate I noticed that the `WeaviateClass` can only be imported from deep inside the package. And then aren't available to import due to the [export rules](https://github.com/weaviate/typescript-client/blob/main/package.json#L7)

```ts
import { WeaviateClass } from "weaviate-client/dist/node/esm/openapi/types.js";
import { WeaviateClass } from "weaviate-client/dist/node/cjs/openapi/types.js";
```

This type is exposed on the `createFromSchema` method of the Collections interface.

https://github.com/weaviate/typescript-client/blob/main/src/collections/index.ts#L153

Please export public types, we have had to mirror this type in our own code base - which could allow for drift we don't mean.

TY

tsmith023:
Closes https://github.com/weaviate/typescript-client/issues/404

### DIFF
--- a/src/collections/index.ts
+++ b/src/collections/index.ts
@@ -176,6 +176,7 @@ export interface Collections {
 }
 
 export default collections;
+export { WeaviateClass } from '../openapi/types.js';
 export * from './aggregate/index.js';
 export * from './backup/index.js';
 export * from './cluster/index.js';


bevzzz:
I can't find the linked issue

bevzzz (approved):

PR #421, opened 2026-04-15
Provide ability to cancel mid-flight searches

### LINKED ISSUE #414: Feature Needed: Allow external AbortSignal for query cancellation in Base.sendWithTimeout()
## Summary

  The underlying `nice-grpc` transport fully supports `AbortSignal` for cancelling in-flight gRPC calls, but the client's `Base.sendWithTimeout()` creates a private `AbortController` for internal timeout management and provides no way to inject an external signal. This means consumers cannot cancel queries when, e.g., the HTTP client disconnects.

  ## Current behavior (`weaviate-client@3.11.0`)

  `grpc/base.js` creates an internal-only controller:

  ```ts
  sendWithTimeout = (send) => {
      const controller = new AbortController();
      const timeoutId = setTimeout(() => controller.abort(), this.timeout * 1000);
      return send(controller.signal).finally(() => clearTimeout(timeoutId));
  };

  The Searcher.call() correctly passes this signal to nice-grpc:

  call = (message) => this.sendWithTimeout((signal) => this.connection
      .search({ ...message }, { metadata: this.metadata, signal, ...retryOptions })
      // ...
  );
```

  So nice-grpc would cancel the gRPC call if the signal aborted — but the only way to trigger that abort is via the internal timeout.
   There is no parameter to pass an external AbortSignal, and SearchOptions / BaseNearOptions in the public API have no signal field.

## Proposed behavior

  Allow consumers to pass an optional AbortSignal that gets composed with the internal timeout signal, so either the timeout or the
  external signal can cancel the call:

```ts
  // Base.sendWithTimeout — proposed change
  sendWithTimeout = (send, externalSignal) => {
      const controller = new AbortController();
      const timeoutId = setTimeout(() => controller.abort(), this.timeout * 1000);

      // Compose: abort on EITHER timeout or external signal
      const composedSignal = externalSignal
          ? AbortSignal.any([controller.signal, externalSignal])
          : controller.signal;

      return send(composedSignal).finally(() => clearTimeout(timeoutId));
  };
```

  At the public API level, query methods could accept an optional signal:

```ts
  const result = await collection.query.nearText('query', {
      limit: 5,
      signal: abortController.signal, // ← new optional field
  });
```

## Use case

  We're building a Fastify + LangGraph AI agent that uses Weaviate for RAG. When the HTTP client disconnects (gateway timeout, user navigates away), we want to cancel in-flight Weaviate queries to free resources. Currently the gRPC call runs to completion (up to 60s query timeout) even after the consumer has moved on.

  Without external signal support, the only workaround is Promise.race with a timeout — which lets the caller move on but leaves the Weaviate gRPC call as a "zombie" that runs until the internal timeout expires.

tsmith023:
Closes https://github.com/weaviate/typescript-client/issues/414

### DIFF
--- a/src/collections/generate/index.ts
+++ b/src/collections/generate/index.ts
@@ -8,6 +8,7 @@ import { toBase64FromMedia } from '../../index.js';
 import { GenerativeSearch } from '../../proto/v1/generative.js';
 import { SearchReply } from '../../proto/v1/search_get.js';
 import { Deserialize } from '../deserialize/index.js';
+import { CallOptions } from '../index.js';
 import { Check } from '../query/check.js';
 import {
   BaseBm25Options,
@@ -74,10 +75,11 @@ class GenerateManager<T, V> implements Generate<T, V> {
 
   public fetchObjects<C extends GenerativeConfigRuntime | undefined = undefined>(
     generate: GenerateOptions<T, C>,
-    opts?: FetchObjectsOptions<T, V>
+    opts?: FetchObjectsOptions<T, V>,
+    callOpts?: CallOptions
   ): Promise<GenerativeReturn<T, V, C>> {
     return Promise.all([
-      this.check.fetchObjects(),
+      this.check.fetchObjects(callOpts),
       this.check.supportForSingleGroupedGenerative(),
       this.check.supportForGenerativeConfigRuntime(generate.config),
     ])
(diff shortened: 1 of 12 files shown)


bevzzz (approved):

src/collections/generate/index.ts
```diff
@@ -8,6 +8,7 @@ import { toBase64FromMedia } from '../../index.js';
 import { GenerativeSearch } from '../../proto/v1/generative.js';
 import { SearchReply } from '../../proto/v1/search_get.js';
 import { Deserialize } from '../deserialize/index.js';
+import { CallOptions } from '../index.js';
```
bevzzz: I can't find `CallOptions` in `src/collections/index.js` file, what am I missing?

tsmith023: Ah it's in `src/collections/query/index.ts` and is re-exported from `src/collections/index.js`. I'll make it more explicit for clarity

PR #422, opened 2026-04-16
Introduce `tokenize` namespace

tsmith023:


src/collections/config/types/index.ts
bevzzz: This should use the new `Stopwords` type

src/tokenize/types.ts
```diff
@@ -0,0 +1,14 @@
+import { Stopwords, Tokenization } from '../collections/types/index.js';
+
+export type TextAnalyzerConfig = {
+  asciiFold?: boolean | { ignore?: string[] };
```
bevzzz:
```suggestion
  asciiFold?: boolean | { ignore: string[] };
```

Shouldn't allow `asciiFold: {}`, right? The client can handle it, but forcing the user to do `asciiFold: true` to enable it seems cleaner.

src/tokenize/types.ts
```diff
@@ -0,0 +1,14 @@
...
+  stopwordPreset?: Stopwords | string;
+};
+
+export type TokenizeResult = {
+  tokenization: Tokenization | string;
+  indexed: string[];
+  query: string[];
+  analyzerConfig?: TextAnalyzerConfig;
+  stopwordConfig?: Partial<Stopwords>;
```
bevzzz: These two fields were removed in the last iteration I think ([ref. Python PR](https://github.com/weaviate/weaviate-python-client/pull/2019/changes#diff-a61b8c690db706c40763f294e884a3947bdc761ce798f480dcc2cb2f609936fbL31-L32))

bevzzz (changes requested):

src/tokenize/types.ts
tsmith023: I wanted `ignore` to be optional in case of future additions of other options under `asciiFold` but the type can actually be relaxed then if we need to anyway so your suggestion is good, will change

src/tokenize/types.ts
tsmith023: tyvm!

bevzzz (approved):

PR #429, opened 2026-04-30
Wire v1.37 tokenization config (textAnalyzer, stopwordPresets) through public types

g-despot:
## Summary

Brings the TS client to parity with the python client for Weaviate v1.37 tokenization config. Pre-patch, users had to fall back to `as any` for per-property `textAnalyzer`, `invertedIndex.stopwordPresets`, and the `/v1/tokenize` `stopwords` / `stopwordPresets` fields.

**Public surface:**
- `TextAnalyzerConfig` — new type used for both per-property `textAnalyzer` and `tokenize.text({ analyzerConfig })`. Ergonomic union: `asciiFold: boolean | { ignore: string[] }`.
- `InvertedIndexConfig.stopwordPresets` — exposed on create / read / update, plus on the `configure.invertedIndex(...)` and `reconfigure.invertedIndex(...)` builders.
- `tokenize.text` — now accepts `stopwords` (one-off block) and `stopwordPresets` (named catalog). Mutually exclusive — passing both rejects client-side with `WeaviateInvalidInputError`. Version-gated at `>= 1.37.2`.

**Schema:** `tools/refresh_schema.sh v1.37.2` refreshed `src/openapi/schema.ts` so `TokenizeRequest` carries `stopwords` (top-level) and the flat `stopwordPresets` shape. CI matrix bumped to `1.37.2`.

## Test plan

- [x] `WEAVIATE_VERSION=1.37.2 npm run test:unit` — 323/323 pass
- [x] `npm run build` / `npm run lint` — clean
- [x] Integration tests against live Weaviate 1.37.2:
  - `test/tokenize/integration.test.ts` — covers `analyzerConfig`, `stopwords` (preset+additions / additions-only / removals-only), `stopwordPresets` (named ref / builtin override), mutex rejection. Inputs/outputs match the python integration suite.
  - `test/collections/tokenization/integration.test.ts` — round-trips `textAnalyzer` and `stopwordPresets` through `collection.config.get()`.


src/tokenize/index.ts
```diff
@@ -21,9 +21,19 @@ const tokenize = (connection: ConnectionGRPC, dbVersionSupport: DbVersionSupport
           )
         );
       }
+      const requiresStopwordsField = opts?.stopwords !== undefined || opts?.stopwordPresets !== undefined;
```
bevzzz: nit: is `requires` the right verb for this? It looks like this variable checks if a parameter has been set

g-despot: good point, it checks if it's present, not required, fixed

src/collections/tokenization/unit.test.ts
```diff
@@ -0,0 +1,68 @@
...
+  it('Server-returned ApiProperty.textAnalyzer can carry every field the public union form expresses', () => {
+    // The server returns the flat OpenAPI shape; the deserializer
+    // (textAnalyzerConfigFromWire) maps it into the public union form. This
+    // assertion exists so that if the OpenAPI shape grows new fields, we
+    // notice and route them through the deserializer.
+    type ApiTextAnalyzer = NonNullable<NonNullable<ApiProperty>['textAnalyzer']>;
+    expectTypeOf<ApiTextAnalyzer>().toHaveProperty('asciiFold');
+    expectTypeOf<ApiTextAnalyzer>().toHaveProperty('asciiFoldIgnore');
+    expectTypeOf<ApiTextAnalyzer>().toHaveProperty('stopwordPreset');
+  });
```
bevzzz: issue: these feel to me a lot like `test("Math works", expect(2+2).equal(4))`, in that they're testing the TS language / type system and not the business logic.

I think the [type gymnastics](https://www.zhenghao.io/posts/type-programming#avoid-type-gymnastics-in-production-code) we sometimes resort to in the package internals are already too complex for a client library. We probably do not need unit tests for type declarations.

src/openapi/schema.ts
```diff
@@ -554,8 +554,6 @@ export interface definitions {
        */
       alias?: string;
     };
-    /** @description resources applicable for MCP actions */
-    mcp?: { [key: string]: unknown };
```
bevzzz, line 558: question: why was this removed?

bevzzz (changes requested):

src/openapi/schema.ts
g-despot: resources have been removed from the MCP permission in v1.37.1 but I forgot to include the change in the TS client, just cleaning it up now

src/collections/tokenization/unit.test.ts
g-despot: Agreed, this is probably a bit too much, removed

src/openapi/schema.ts
bevzzz: Oh, I didn't realize this was an auto-generated file, pardon

test/tokenize/integration.test.ts
```diff
@@ -85,3 +86,71 @@ requireAtLeast(1, 37, 0).describe('tokenize integration test', () => {
     });
   });
 });
+
+requireAtLeast(1, 37, 2).describe('tokenize stopwords / stopwordPresets (>= 1.37.2)', () => {
```
bevzzz:
```suggestion
requireAtLeast(1, 37, 2).describe('tokenize stopwords / stopwordPresets', () => {
```

nit: the clarification is redundant as `requireAtLeast(1, 37, 2)` says the same

bevzzz (approved):

src/collections/config/utils.ts
```diff
@@ -63,6 +64,52 @@ import {
...
+  };
+  if (typeof config.asciiFold === 'boolean') {
+    out.asciiFold = config.asciiFold;
+  } else if (typeof config.asciiFold === 'object') {
+    out.asciiFold = true;
+    out.asciiFoldIgnore = config.asciiFold.ignore;
+  }
+  return out.asciiFold !== undefined || out.asciiFoldIgnore !== undefined || out.stopwordPreset !== undefined
+    ? out
+    : undefined;
```
bevzzz:
```suggestion
  return out;
```

nit: is there any harm in returning an empty `{}` object if all it's properties are `?`-optional? Especially on the wire, where these will be turned to nulls or omitted entirely.
The return `long_condition ? out : undefined seems unnecessary`.

PR #432, opened 2026-05-21
Enhance vector index handling and testing

mpartipilo:
Starting with Weaviate 1.37.5 the server sets a default vector index type to set when the client does not specify one. This PR establishes that no vector index type is set by default, and only sets "hsnw" as default for servers running <=1.37.4.

PR #433, opened 2026-05-22
Follow-up: address remaining default vector index review fixes

mpartipilo:
This follow-up PR contains changes that were committed after PR #432 was merged.

Included:
- rename abbreviation-heavy identifiers in tests/code for clarity
- move architecture-suffix version parsing into DbVersion.fromString
- add regression coverage for architecture-suffixed versions
- simplify integration test default expectation and related comments

Context:
- Original PR: #432
- Missing commit: 40568003c3a61238d6aed8f23f0fbb4d2dcec672


test/collections/defaultVectorIndexType/integration.test.ts
```diff
@@ -7,27 +7,19 @@ import { DbVersion } from '../../../src/utils/dbVersion.js';
...
+const parsedVersion = DbVersion.fromString(`v${WEAVIATE_VERSION}`);
 
 // >= 1.37.5: server applies DEFAULT_VECTOR_INDEX itself; client must NOT inject.
 // <  1.37.5: server has no such feature; client injects vectorIndexType = 'hnsw'.
 const serverAppliesDefault = parsedVersion.isAtLeast(1, 37, 5);
 
-// Use a linux/amd64 platform pin only when the image tag carries the ".amd64"
-// suffix — same narrowing the old test did, now applied to a single version.
+// Use a linux/amd64 platform pin only when the image tag carries the ".amd64" suffix.
 const platform = WEAVIATE_VERSION.includes('.amd64') ? 'linux/amd64' : undefined;
```
bevzzz: suggestion: iiuc, this is used as a parameter to `startContainer`. Could that function handle this piece of logic then? The benefit of it being that every caller of `startContainer` will not need to consider platform pin individually.

test/collections/defaultVectorIndexType/integration.test.ts
```diff
@@ -7,27 +7,19 @@ import { DbVersion } from '../../../src/utils/dbVersion.js';
...
 // Version resolution
 //
-// WEAVIATE_VERSION follows the existing convention (see test/version.ts) and
-// is required — CI sets it; locally export it before running this test.
 // ─────────────────────────────────────────────────────────────────────────────
 const WEAVIATE_VERSION = process.env.WEAVIATE_VERSION;
 if (!WEAVIATE_VERSION) {
   throw new Error('WEAVIATE_VERSION env var is required for this integration test');
 }
-const expectedDefault = process.env.DEFAULT_VECTOR_INDEX ?? 'hfresh';
```
bevzzz: question: could this test run against one of the containers spinned up by `./ci/run_containers.sh`? My understanding is most integration tests run against those, so it'd be good to keep the harnessing consistent

mpartipilo: Yeah, this is a leftover from when I used testcontainers to run before/after 1.37.5. I'll make it uniform with the rest of the tests.

test/collections/defaultVectorIndexType/integration.test.ts
mpartipilo: This may become unnecessary and I'll end up deleting it

test/collections/defaultVectorIndexType/integration.test.ts
```diff
@@ -1,97 +1,46 @@
...
  * Assert that a collection's stored index type matches what the server/client
  * should have applied given the running version:
- *   - server >= 1.37.5 → server applied DEFAULT_VECTOR_INDEX (e.g. 'hfresh')
- *   - server <  1.37.5 → client injected 'hnsw' as legacy default
+ *   - on newer servers, expect hfresh default
+ *   - on older servers, expect legacy hnsw default
  */
 function assertDefaultIndexType(actual: string) {
-  expect(actual).toEqual(serverAppliesDefault ? expectedDefault : 'hnsw');
+  expect(actual).toEqual(expectedDefaultIndexType(serverVersion));
```
bevzzz: nit: do we need a helper function for a ternary operator? (`expectedDefaultIndexType`)

test/collections/defaultVectorIndexType/integration.test.ts
```diff
@@ -1,97 +1,46 @@
...
-// DbVersion.fromString understands semver pre-release labels (e.g. -e0fe0d5)
-// but not dot-separated platform tokens appended after them.
-const versionForParsing = WEAVIATE_VERSION.replace(/\.(amd64|arm64|x86_64)$/, '');
-const parsedVersion = DbVersion.fromString(`v${versionForParsing}`);
-
-// >= 1.37.5: server applies DEFAULT_VECTOR_INDEX itself; client must NOT inject.
-// <  1.37.5: server has no such feature; client injects vectorIndexType = 'hnsw'.
-const serverAppliesDefault = parsedVersion.isAtLeast(1, 37, 5);
+let client: WeaviateClient;
+let serverVersion: DbVersion;
```
bevzzz: nit: imo these are better scoped to `describe()`, not the module, considering their lifecycle is managed by `beforeAll` and `afterAll` hooks

src/collections/configure/unit.test.ts
```diff
@@ -1258,10 +1258,8 @@ describe('Unit testing of the vectorizer factory class', () => {
     const config = configure.vectors.text2VecDigitalOcean({ model: 'qwen3-embedding-0.6b' });
     expect(config).toEqual<VectorConfigCreate<never, undefined, 'hnsw', 'text2vec-digitalocean'>>({
       name: undefined,
-      vectorIndex: {
-        name: 'hnsw',
-        config: undefined,
-      },
+      properties: undefined,
+      vectorIndex: undefined,
```
bevzzz: question: do these need to be explicitly set to `undefined`? Do they default to something different if left unset?

mpartipilo: I am following the pattern of the other tests now.

test/collections/config/integration.test.ts
```diff
@@ -119,26 +139,35 @@ describe('Testing of the collection.config namespace', () => {
...
-        : 'acorn',
-      flatSearchCutoff: 40000,
-      distance: 'cosine',
-      multiVector: undefined,
-      quantizer: undefined,
-      type: 'hnsw',
-    });
-    expect(config.vectorizers.default.indexType).toEqual('hnsw');
+    const expectedIndexType2 = expectedDefaultIndexType(serverVersion);
+    if (expectedIndexType2 === 'hfresh') {
```
bevzzz: nit: please remove the redundant qualifier `2`. There's no `expectedIndexType` in this scope as far as I can tell

test/collections/config/integration.test.ts
```diff
@@ -61,26 +72,35 @@ describe('Testing of the collection.config namespace', () => {
...
+        dynamicEfMin: 100,
+        dynamicEfMax: 500,
+        dynamicEfFactor: 8,
+        vectorCacheMaxObjects: 1000000000000,
+        filterStrategy: serverVersion.isLowerThan(1, 34, 0) ? 'sweeping' : 'acorn',
+        flatSearchCutoff: 40000,
+        distance: 'cosine',
+        multiVector: undefined,
+        quantizer: undefined,
+        type: 'hnsw',
```
bevzzz: issue: I think making this assertion any more detailed than `expect index is HNSW` or `expect index is HFresh` makes this test more brittle, i.e if the next version of Weaviate uses `dynamicEfMin = 104` this is going to break. 

suggestion: avoid asserting on values controlled by the server

test/collections/config/integration.test.ts
```diff
@@ -15,13 +15,24 @@ import weaviate, {
...
+import { DbVersion } from '../../../src/utils/dbVersion.js';
 import { requireAtLeast } from '../../../test/version.js';
 
+/**
+ * Helper to determine expected default vector index type based on server version.
+ * On versions >= 1.37.5 with DEFAULT_VECTOR_INDEX=hfresh, the server defaults to hfresh.
+ */
+function expectedDefaultIndexType(version: DbVersion): 'hnsw' | 'hfresh' {
+  return version.isAtLeast(1, 37, 5) ? 'hfresh' : 'hnsw';
+}
```
bevzzz: nit: this doesn't need to be in a function if it's only used twice and can be inlined similarly to how `maxConnections: serverVersion.isLowerThan(1, 26, 0) ? 64 : 32` is

bevzzz (dismissed):

test/collections/config/integration.test.ts
mpartipilo: Done — inlined the ternary at the 4 call sites and removed the helper in 09cc3a3.

test/collections/config/integration.test.ts
mpartipilo: Good point — narrowed both default-index assertions to just check `indexType` is `hnsw` or `hfresh` based on server version (09cc3a3). Dropped the brittle field-by-field config matchers.

test/collections/config/integration.test.ts
mpartipilo: Renamed — `expectedIndexType2` is gone (now inlined alongside the other fix in 09cc3a3).

bevzzz (approved):

PR #447, opened 2026-06-19
feat: add optional location parameter to text2vec-google vectorizer

### LINKED ISSUE #442: Add support for location parameter in text2vec-google module
Add support for location parameter in text2vec-google module. This is an optional parameter.

g-despot:
Adds an optional `location` parameter (Google Vertex AI region) to the `text2vec-google` vectorizer config. When unset it is omitted from the module config, so the server applies its default.

Closes #442

### DIFF
--- a/src/collections/configure/unit.test.ts
+++ b/src/collections/configure/unit.test.ts
@@ -1410,6 +1410,7 @@ describe('Unit testing of the vectorizer factory class', () => {
     const config = configure.vectors.text2VecGoogle({
       name: 'test',
       apiEndpoint: 'api-endpoint',
+      location: 'us-central1',
       model: 'model-id',
       projectId: 'project-id',
     });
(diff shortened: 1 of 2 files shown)


src/collections/configure/unit.test.ts
```diff
@@ -1427,6 +1429,39 @@ describe('Unit testing of the vectorizer factory class', () => {
...
+    expect(withoutLocation).toEqual<VectorConfigCreate<never, undefined, 'hnsw', 'text2vec-google'>>({
+      name: undefined,
+      vectorizer: {
+        name: 'text2vec-google',
+        config: {
+          projectId: 'project-id',
+        },
+      },
+    });
+  });
```
bevzzz: Seems to me the test above, i.e `'Unit testing of the vectorizer factory class'` covers this sufficiently and this one is perhaps redundant? Looking at the actual changes in `vectorizer.ts` I doubt this test can even fail.

g-despot: You're right on both points, dropped it

bevzzz (approved):

PR #448, opened 2026-06-22
feat: add optional dimensions to text2vec-aws vectorizer

### LINKED ISSUE #443: Add support for dimensions setting in text2vec-aws module
Add support for optional dimensions setting in text2vec-aws module. This is an integer value, defaults to nil in Weaviate.

g-despot:
Adds an optional `dimensions` parameter (output embedding dimensionality) to the `text2vec-aws` vectorizer config. Omitted when unset, so the server applies its default.

Closes #443


bevzzz (approved):
Good one!

PR #449, opened 2026-06-26
feat(query): Support diversity selection for near- queries

bevzzz:


test/collections/query/integration.test.ts
```diff
@@ -182,6 +183,13 @@ describe('Testing of the collection.query methods with a simple collection', ()
     });
   });
 
+  requireAtLeast(1, 37, 8).it('near text query with diversity selection', async () => {
+    const ret = await collection.query.nearText(['carrot'], {
+      diversity: Diversity.mmr({ limit: 1 }),
+    });
+    expect(ret.objects).toHaveLength(1);
+  });
```
tsmith023, line 199: wdyt about duplicating this test to show usage without the factory method to ensure it's an accepted journey wrt the type system? e.g.
```typescript
diversity: {
  type: 'mmr',
  limit: 1,
}
```

bevzzz: Added 👌 

nb: I don't know that there are (m)any other tests which show usage without the factories for entities like HNSW index config or different vectorizers even though conceptually that can be a user's preferred approach to write code.

tsmith023: ✅ I reckon that's an oversight on my part when developing these tests previously

tsmith023 (approved):

PR #450, opened 2026-06-26
feat(query): add Boost parameter to all query types

### LINKED ISSUE #435: replicationConfig.asyncEnabled is being removed in Weaviate v1.38 — drop client-side field
## Background

Weaviate is removing the `replicationConfig.asyncEnabled` field from collection schemas. This is the server-side commit https://github.com/weaviate/weaviate/commit/cef789345e (part of [weaviate/weaviate#11214](https://github.com/weaviate/weaviate/pull/11214)). After it ships, the field will:

- be silently dropped when present in `POST /v1/schema` request bodies (Go's `encoding/json` ignores unknown fields)
- not appear in `GET /v1/schema` response bodies

The post-removal behaviour is governed by the existing global `ASYNC_REPLICATION_DISABLED` runtime override: for any class with `factor > 1`, async replication runs by default unless that flag is set.

## Impact on `weaviate-client` (latest: v3.13.0)

**Technically safe — no crash.** The read path defaults missing fields to `false`:

https://github.com/weaviate/typescript-client/blob/main/src/collections/config/utils.ts#L485

```ts
asyncEnabled: v.asyncEnabled ? v.asyncEnabled : false,
```

**But the value the client reports back is misleading.** Every `replicationConfig.asyncEnabled` returned by `collection.config.get()` will be `false`, even when the server is actually running async replication (the new default for any class with `factor > 1`).

The write path (`src/collections/configure/index.ts` L189, L195, L201, L315, L321) still sends `asyncEnabled` to the server. The server silently drops it, so user intent is lost without any indication.

## Recommended change

1. Remove `asyncEnabled` from the read-side `ReplicationConfig` type at `src/collections/config/types/index.ts:77`.
2. Drop the default-to-false branch in `src/collections/config/utils.ts:485`.
3. Remove `asyncEnabled` from the write-side configure options in `src/collections/configure/index.ts` (the parameter, the JSDoc, and the merge into the outgoing payload at L189–321).

The server change is backward compatible at the wire level (existing classes with `asyncEnabled` persisted in their schemas are read fine — the field is dropped, not validated). So clients that don't ship the change still work; they just report a misleading `false`.

## Target server version

Weaviate v1.38 (the release that includes #11214).

cc @weaviate/typescript-client-maintainers

### LINKED ISSUE #440: Boost API
> Mirrored from weaviate/weaviate-python-client#2068 (https://github.com/weaviate/weaviate-python-client/issues/2068)

PRs for reference: 
- code: https://github.com/weaviate/weaviate-python-client/pull/2030
- docstrings: https://github.com/weaviate/weaviate-python-client/pull/2057

bevzzz:
Closes #440 
Closes #435


trengrj (approved):

trengrj (approved):

anschm:
Hello, thanks for the implementation of the new boost feature. Could we get this in in the next release? We are using the weavaite typescript client and we would like to use the new boost feature in our product. Thank you.

bevzzz:
hey @anschm! Thanks for reaching out -- I'm just waiting for the pipeline to go green and then we'll be making a new release right away.

bevzzz:
The [latest release](https://github.com/weaviate/typescript-client/releases/tag/v3.14.0) is live. Check it out, @anschm 😉

PR #459, opened 2026-08-03
Add incremental backup support

dudanogueira:
Exposes Weaviate's file-based incremental backups (server `v1.37.0`+) in the client:

```ts
const base = await client.backup.create({
  backupId: 'base-backup',
  backend: 'filesystem',
  waitForCompletion: true,
});

await client.backup.create({
  backupId: 'incremental-backup',
  backend: 'filesystem',
  config: { incrementalBaseBackupId: base.id },
  waitForCompletion: true,
});
```

Unchanged files are not copied — they are restored from the base. Deleting a base backup therefore breaks every incremental built on top of it.

### Changes

- `incrementalBaseBackupId` on `BackupConfigCreate`, so it is available on `client.backup.create()` and `collection.backup.create()`, plus `.withIncrementalBaseBackupId()` on the v2 builder.
- Gated by `DbVersionSupport.supportsIncrementalBackups()` — throws `WeaviateUnsupportedFeatureError` below `1.37.0`, matching the Python client. The gate applies to the v2 builder too: Weaviate below `1.37.0` silently ignores the field and writes a *full* backup, which is worse than an error.
- The base backup ID is read back on `getCreateStatus()` and `list()`.
- `parseStatus()` now also surfaces `size`, which was declared on `BackupStatusReturn` but never populated.

### When `incrementalBaseBackupId` comes back as `undefined`

The read-back has a narrower floor than the feature itself, so it is `undefined` in more cases than you might expect: when the backup is not incremental, when the caller is not an RBAC **root** user, when the server is older than **`v1.37.6`** (the create-side field landed much earlier; the status/list read-back was added in `7a53058`), and on the return of `create()` without `waitForCompletion`, since `BackupCreateResponse` carries no such field.

### Notes for reviewers

- `backup(connection)` now takes `dbVersionSupport` as a second argument — the v3 factory, `backupCollection`, and the v2 factory in `src/backup/index.ts`. All three are internal; only types are re-exported publicly. `BackupCreator`'s new constructor argument is optional, so the public class stays source-compatible.
- `collection.backup.create()` accepts `config` now; `restore` is unchanged.
- `list()` no longer returns the raw payload verbatim: `incremental_base_backup_id` is mapped to `incrementalBaseBackupId`. Everything else is passed through unchanged.

### Testing

Two integration tests: one creates an incremental on top of a base and restores it, and one asserts the base backup ID reaches the server by pointing it at a base that does not exist and expecting the server's `could not fetch base backup` error.

The `<1.37.0` version gate has no automated test — it can only be reached against a server reporting an old version. Verified by hand against 1.36.0 on both the v3 and v2 paths. Feature verified against 1.38.2.

https://claude.ai/code/session_014t3DrgYbqzyQyiKjQ1UfwW


src/utils/dbVersion.ts
```diff
@@ -163,6 +163,13 @@ export class DbVersionSupport {
       message: this.errorMessage('Tokenize endpoint stopwords / stopwordPresets', version.show(), '1.37.2'),
     }));
 
+  supportsIncrementalBackups = () =>
+    this.dbVersionProvider.getVersion().then((version) => ({
+      version,
+      supports: version.isAtLeast(1, 37, 0),
```
g-despot, line 169: 1.37.0 is too high. I ran 1.36.10 and the server accepted the field and produced a genuine incremental. Tag scan says `create` landed in 1.35.13 and 1.36.3. `supportsSingleGrouped` just above has the pattern for backported ranges.

Worth splitting this into two checks, since the read-back has a different floor (~1.37.6).

src/backup/backupCreator.ts
```diff
@@ -70,12 +72,27 @@ export default class BackupCreator extends CommandBase {
...
+  /**
+   * The ID of an existing backup to use as the base for a file-based incremental backup.
+   * Only the files that changed since the base backup are included in the new backup.
+   *
+   * `backupId` is a plain backup ID string: either a literal, e.g. `'my-base-backup'`, or the
+   * `id` returned by a previous backup creation.
+   *
+   * Requires Weaviate v1.37.0 or higher.
+   */
+  withIncrementalBaseBackupId(backupId: string) {
```
g-despot, line 86: This path skips both the version gate and the lowercasing that `client.ts` does, and `validate()` at :95 compares raw, so `backupId: 'B1'` with base `'b1'` gets through. `weaviateV2` is public API.

Lowercasing here would fix the validate case for free.

src/collections/backup/client.ts
```diff
@@ -47,6 +54,9 @@ export const backup = (connection: Connection): Backup => {
       error: res.error,
       path: res.path,
       status: res.status,
+      // Only returned by Weaviate >=1.37, for incremental backups, and only to root users
```
g-despot: While you're in this object: `size` is declared on `BackupStatusReturn` but never set here, so it's always undefined. The server does return it and Python surfaces it. One line?

src/collections/backup/types.ts
```diff
@@ -16,6 +16,8 @@ export type BackupStatusReturn = {
   status: BackupStatus;
   /** Size of the backup in Gibs */
   size?: number;
+  /** The ID of the base backup this incremental backup was built on; undefined if the backup is not incremental. */
```
g-despot: It's also undefined for non-root users, on servers below ~1.37.6, and always when `waitForCompletion` is false, since `BackupCreateResponse` doesn't carry the field at all. Worth spelling out. This is exactly the field where a wrong comment sends someone hunting the wrong bug.

src/collections/backup/types.ts
```diff
@@ -64,6 +66,20 @@ export type BackupArgs<C extends BackupConfigCreate | BackupConfigRestore> = {
   config?: C;
 };
 
+/** The arguments required to create a backup. */
+export type BackupCreateArgs = BackupArgs<BackupConfigCreate> & {
+  /**
```
g-despot: Can we get this down to one line? The rest of the file is one-liners and this block is repeated near-verbatim in `collection.ts` and `backupCreator.ts`.

Worth stealing the sentence from the Python docstring though: unchanged files are restored from the base, so deleting a base backup breaks every incremental built on it. That's the part people actually need.

test/collections/backup/integration.test.ts
```diff
@@ -258,6 +258,65 @@ describe('Integration testing of backups', () => {
...
+        backupId: randomBackupId(),
+        backend: 'filesystem',
+        includeCollections: [collection.name],
+        incrementalBaseBackupId: base.id,
+        waitForCompletion: true,
+      });
+      expect(incremental.status).toBe('SUCCESS');
+
+      // Weaviate only reports the base backup ID to root users, so treat it as optional
+      if (incremental.incrementalBaseBackupId !== undefined) {
```
g-despot: This never runs. The server only returns `incremental_base_backup_id` to RBAC root users, and `docker-compose-backup.yml` is anonymous, so the guard is always false. (Also `tests-with-auth` is skipped for fork PRs.)

I did verify it works with a root user on 1.38.9, all three surfaces round-trip correctly. Either wire up a root case or drop the `if` so the coverage is honest.

g-despot (changes requested):
Tested this locally against 1.35.0, 1.36.10 and 1.38.9.

The feature works. I confirmed the incremental genuinely skips base segments on disk (1.7 MB vs 35 MB uncompressed, with the objects and property_text segments absent), and that restoring the incremental fails without the base present and succeeds with it.

Two things I'd like sorted before merge:

1. The 1.37.0 floor is too high. 1.36.10 accepts `incremental_base_backup_id` and produces a real incremental (345 KB vs 7.5 MB base, `baseBackupId` persisted in `backup_config.json`), but the client throws `WeaviateUnsupportedFeatureError` against it. A tag scan puts `create` at 1.35.13 and 1.36.3. Below that I checked 1.35.0, where the server silently ignores the field and you get a full backup, so the gate is worth keeping, just lower.

2. The read-back is a separate feature with its own floor. `incremental_base_backup_id` only comes back on status/list from ~1.37.6, and only for root users. On 1.37.0 to 1.37.5 we'd be promising a field the server never sends.

Rest inline. Nice work otherwise, the design is clean.

src/utils/dbVersion.ts
g-despot: Actually maybe we can ignore this one, as I see that Python also uses 1.37.0

dudanogueira: Taking your follow-up — keeping the floor at 1.37.0 so we stay in step with Python (`backup/executor.py` gates on `is_lower_than(1, 37, 0)`).

Your tag scan checks out on my side too: the incremental-create merge `bb6b4be` first appears in v1.34.18 / v1.35.13 / v1.36.3. So the lower floor is real, it just feels like a coordinated change across clients rather than TS diverging on its own. Happy to open a follow-up issue for that if you want it tracked.

src/backup/backupCreator.ts
dudanogueira: Fixed both halves.

`withIncrementalBaseBackupId` now lowercases, and `validateIncrementalBaseBackupId` compares case-insensitively. That second part matters for the exact case you named: lowercasing the base alone would not catch `backupId: 'B1'` + base `'b1'`, because the primary ID keeps its casing on the v2 path.

I also added the version gate rather than leaving it to v3. `BackupCreator` takes an optional `DbVersionSupport` and `do()` checks it before posting; `weaviateV2.client()` passes it through. Worth having given you confirmed 1.35.0 silently produces a full backup — a silent full backup is exactly the failure the gate exists to prevent. Mock coverage under `with the v2 builder` and `should throw from the v2 builder too`.

The v3 path now delegates the lowercasing to the builder instead of doing its own.

src/collections/backup/client.ts
dudanogueira: Done — `size: 'size' in res ? res.size : undefined`. The `in` guard is needed because `parseStatus` also handles restore responses, which have no `size` field. Mock asserts it now.

src/collections/backup/types.ts
dudanogueira: Rewritten to spell out all four cases: not incremental, non-root caller, server older than 1.37.6, and the return of `create()` without `waitForCompletion`.

Confirmed the 1.37.6 number from the server repo rather than guessing — `7a53058` ("Add root only baseBackupId return") first appears in tag v1.37.6. Good catch, thanks.

src/collections/backup/types.ts
dudanogueira: One line now, and I stole the Python sentence — unchanged files are restored from the base, so deleting a base backup breaks every incremental built on it.

Also removed the duplicates you were pointing at: `BackupCollectionCreateArgs` is now `BackupCollectionArgs & Pick<BackupCreateArgs, 'incrementalBaseBackupId'>`, and the two interface JSDoc blocks in `client.ts` / `collection.ts` are one line each.

test/collections/backup/integration.test.ts
dudanogueira: Did both, since neither alone is honest on its own.

The anonymous suite now asserts the field **is** undefined — that is the real contract there, and it would have caught a regression that started leaking the ID to non-root callers.

For the root case: `AUTHORIZATION_ADMIN_USERS` populates the RBAC root-user list server-side (`usecases/config/environment.go` falls back to it when `AUTHORIZATION_RBAC_ROOT_USERS` is unset), so `admin-user` on `ci/docker-compose-rbac.yml` is already a root caller. I added `backup-filesystem` + `BACKUP_FILESYSTEM_PATH` to that compose and a new `requireAtLeast(1, 37, 6)` suite on 8091 with `ApiKey('admin-key')` that asserts the round-trip across `create`, `getCreateStatus` and `list`.

`compose_up_all` brings that file up in every job and it needs no secrets, so it runs on fork PRs too — it just will not execute anywhere below 1.37.6, and CI's `WEAVIATE_137` is 1.37.5, so in practice only the 1.38 legs exercise it.

Verified locally against 1.38.2: all three surfaces round-trip.

dudanogueira:
Pushed the review fixes in f493925a. Summary of what changed per thread is inline; the PR description is rewritten too.

### CI

Two legs are red, and I want to be upfront that I looked into whether I caused them. I do not believe I did:

- `tests-without-auth (22.x, 1.38)` — `collection.query > boost parameter > 'timeDecay'`, an ordering assertion. Unrelated to backups.
- `tests-without-auth (24.x, 1.37)` — two `test/users/integration.test.ts` role assign/revoke 404s, plus a `deleteMany` "Stringified UUID is invalid".

The two users failures are a known flake: **the identical pair failed on `main` on 2026-08-13** (run 31690118238, on the 1.34 leg), with no backup code in play.

The one leg that actually exercises the new root-user suite is `tests-without-auth (24.x, 1.38)`, because `requireAtLeast(1, 37, 6)` skips it everywhere else and CI's `WEAVIATE_137` is 1.37.5. **That leg passed.**

The only way my change could plausibly touch the users/roles suites is the new `backup-filesystem` module on `ci/docker-compose-rbac.yml`, so I ran both suites locally against that exact modified compose on 1.38.2: 39/39 pass.

I do not have rerun rights on this repo — if someone with admin could kick the two failed jobs, that should close it out.

test/collections/backup/mock.test.ts
```diff
@@ -186,3 +192,236 @@ describe('Mock testing of backup cancellation', () => {
...
+        .withBackupId(BACKUP_ID)
+        .withBackend(BACKEND)
+        .withIncrementalBaseBackupId(BASE_BACKUP_ID)
+        .do();
+      await expect(promise).rejects.toThrow(WeaviateUnsupportedFeatureError);
+    });
+
+    afterAll(() => mock.close());
+  });
+});
```
bevzzz: This PR already adds e2e test for this feature, so I really don't think there is a need for 250 lines of mock tests. Could we please remove these altogether?

src/backup/validation.ts
```diff
@@ -45,3 +45,19 @@ export function validateBackupId(backupId?: string) {
...
+  if (!isValidStringProperty(incrementalBaseBackupId)) {
+    return [
+      'string incrementalBaseBackupId must be a non-empty string - set with .withIncrementalBaseBackupId(backupId)',
+    ];
+  }
+  // Weaviate treats backup IDs as case-insensitive, so 'B1' and 'b1' name the same backup.
+  if (incrementalBaseBackupId.toLowerCase() === backupId?.toLowerCase()) {
+    return ['incrementalBaseBackupId must be different from the ID of the backup being created'];
+  }
+  return [];
```
bevzzz: issue: This validation should not be in the client, it is entirely up to the server to resolve backup IDs and allow/forbid them.

src/collections/backup/client.ts
```diff
@@ -47,6 +54,11 @@ export const backup = (connection: Connection): Backup => {
       error: res.error,
       path: res.path,
       status: res.status,
+      // Restore responses carry neither of these fields; see BackupStatusReturn for when Weaviate
+      // omits `incremental_base_backup_id` from a create status.
+      size: 'size' in res ? res.size : undefined,
+      incrementalBaseBackupId:
+        'incremental_base_backup_id' in res ? res.incremental_base_backup_id || undefined : undefined,
     };
```
bevzzz: issue: these checks are entirely unnecessary. If a key is not present in the `res` object then the expression already invalidates to `undefined`.

See [playground example](https://www.typescriptlang.org/play/?#code/DYUwLgBATiDOBcEBKIDGB7KATAPLMUAlgHYDmANBAIbECeAfBALwQDeADlWABaIBEAek48BYdAIBGVVAGsAruz6VCWflJkBaAIx8AvgG4AUIYzFY6UADpg6UgAoYsS7EIAvEAEojp81Zv2Achd3AIgSaDgIAH4Ip2CQCAAfRIg5YiwQADMSECwIRDSM7OJcr0MgA).

src/collections/backup/types.ts
```diff
@@ -64,6 +72,12 @@ export type BackupArgs<C extends BackupConfigCreate | BackupConfigRestore> = {
   config?: C;
 };
 
+/** The arguments required to create a backup. */
+export type BackupCreateArgs = BackupArgs<BackupConfigCreate> & {
+  /** The ID of an existing backup to build a file-based incremental backup on. Files identical to the base are not copied and are restored from the base instead, so deleting a base backup breaks every incremental built on it. Requires Weaviate `v1.37.0` or higher. */
+  incrementalBaseBackupId?: string;
+};
+
```
bevzzz: issue: why is it necessary to declare a new type? `BackupConfigCreate` can just be extended with `incrementalBaseBackupId`.

ci/docker-compose-rbac.yml
```diff
@@ -16,7 +16,11 @@ services:
       - 50062:50051
     restart: on-failure:0
     environment:
-      ENABLE_MODULES: "generative-dummy,reranker-dummy"
+      # backup-filesystem is enabled here so that backup behaviour which Weaviate only exposes to
+      # root users (e.g. incremental_base_backup_id) can be tested; ci/docker-compose-backup.yml
+      # is anonymous and therefore never has a root caller.
```
bevzzz: nit: redundant comment, please remove

test/collections/backup/integration.test.ts
```diff
@@ -258,7 +258,119 @@ describe('Integration testing of backups', () => {
...
+        backupId: randomBackupId(),
+        backend: 'filesystem',
+        includeCollections: [collection.name],
+        incrementalBaseBackupId: base.id,
+        waitForCompletion: true,
+      });
+      expect(incremental.status).toBe('SUCCESS');
+
+      // This instance is anonymous, so the caller is never a root user and Weaviate withholds the
+      // base backup ID. The root-user round-trip is covered by the RBAC suite below.
```
bevzzz: nit: please remove comment slop

test/collections/backup/integration.test.ts
```diff
@@ -258,7 +258,119 @@ describe('Integration testing of backups', () => {
...
+      backupId: incremental.id,
+      backend: 'filesystem',
+    });
+    expect(status.incrementalBaseBackupId).toBe(base.id);
+
+    const listed = await client.backup.list('filesystem');
+    expect(listed.find((b) => b.id === incremental.id)?.incrementalBaseBackupId).toBe(base.id);
+    expect(listed.find((b) => b.id === base.id)?.incrementalBaseBackupId).toBeUndefined();
+  });
+});
```
bevzzz: issue: `Weaviate only returns incremental_base_backup_id to root users, and only from v1.37.6 onwards` is a property _of Weaviate server_ and not of the client. Please remove this test.

ci/docker-compose-rbac.yml
```diff
@@ -16,7 +16,11 @@ services:
...
+      ENABLE_MODULES: "generative-dummy,reranker-dummy,backup-filesystem"
+      BACKUP_FILESYSTEM_PATH: "/tmp/backups"
```
bevzzz: > backup-filesystem is enabled here so that backup behaviour which Weaviate only exposes root users (e.g. incremental_base_backup_id) can be tested

This is not an acceptance test suite for the Weaviate server and is not aimed at testing the server's behaviour. Please revert this change and the associated `describe` block in `test/collections/backup/integration.test.ts`

bevzzz (changes requested):
"Incremental backup support" is 1 string field in the request payload and 1 string field in the response payload. It is surprising to see that the PR is over 500LOC.

We want to have a test or two to check that the values reach the server, which will add another 100-150 lines. But that's about it. Adding 300 lines of redundant comments / tests with each PR is simply unsustainable and it makes the PR that much more difficult to review.

Please take care to minimize the size of your contributions to changes that are strictly necessary.

src/backup/validation.ts
dudanogueira: Removed. `validateIncrementalBaseBackupId` is gone and `src/backup/validation.ts` is back to its state on `main`.

I also dropped the lowercasing that went with it (`withIncrementalBaseBackupId` now stores the ID as given). Same reasoning — the server resolves backup IDs, so the client should not be normalising them on the way out.

src/collections/backup/client.ts
dudanogueira: Removed. `parseStatus` now reads `res.size` and `res.incremental_base_backup_id` directly.

One note on why the guards were there: `parseStatus` took `BackupCreateStatusResponse | BackupRestoreResponse`, and `BackupRestoreResponse` declares neither field, so `tsc` rejected the plain access on the union. Changing the parameter to an intersection fixes it — every field is optional, so both response types still assign — and the body is now what you asked for.

src/collections/backup/types.ts
dudanogueira: Done — `BackupCreateArgs` and `BackupCollectionCreateArgs` are both gone, `incrementalBaseBackupId` is a field on `BackupConfigCreate`, and `create` is back to `BackupArgs<BackupConfigCreate>`.

Two consequences worth flagging, since you may want them called differently:

1. It changes the call shape to `create({ backupId, backend, config: { incrementalBaseBackupId } })`. The Python client has it as a top-level argument next to `config` (`weaviate/backup/sync.pyi`), matching the wire payload, where `incremental_base_backup_id` is a sibling of `config` rather than inside it. So this diverges from Python. Happy to move it back out if parity matters more than the type count — it can be inlined on the signature without declaring a named type.
2. `BackupCollectionArgs` is shared by `create` and `restore` and had no `config` at all, so adding one there would have put a create config on `restore`. I left the type alone and widened only the create signature to `BackupCollectionArgs & { config?: BackupConfigCreate }`.

test/collections/backup/mock.test.ts
dudanogueira: Removed — `test/collections/backup/mock.test.ts` is back to its state on `main`.

The integration side is now two tests: one creates an incremental on top of a base and restores it, and one asserts the base backup ID actually reaches the server by pointing it at a base that does not exist and expecting the server error (`could not fetch base backup`). That second one is what gives us the "values reach the server" coverage without a mock — the read-back could not, since Weaviate only returns the field to root users.

One thing that is now uncovered: the `<1.37.0` version gate. It can only be exercised against a server that reports an old version, so no integration test can reach it. I verified both paths by hand against 1.36.0 and they throw `WeaviateUnsupportedFeatureError`. Say the word if you would rather the gate went too — it is about 30 lines across `BackupCreator` and `DbVersionSupport`.

test/collections/backup/integration.test.ts
dudanogueira: Done.

test/collections/backup/integration.test.ts
dudanogueira: Removed — the whole `Integration testing of incremental backups as a root user` block is gone.

ci/docker-compose-rbac.yml
dudanogueira: Reverted — `ci/docker-compose-rbac.yml` is back to its state on `main`, and the `describe` block that needed it is gone.

src/backup/backupCreator.ts
dudanogueira: @g-despot heads up — this one is reverted in `6395914c`. @bevzzz asked for the client-side check to go ([here](https://github.com/weaviate/typescript-client/pull/459#discussion_r3795524025)): resolving backup IDs is the server's job, not the client's. So the lowercasing and the self-reference check are both gone, and `backupId: 'B1'` with base `'b1'` now reaches the server and gets whatever answer the server gives.

The version gate you argued for is still there, and it now covers the v2 builder too, so the 1.35.0 silent-full-backup case is still an error rather than a surprise.

test/collections/backup/integration.test.ts
dudanogueira: @g-despot heads up — this one is reverted in `6395914c`. @bevzzz asked for the root-user suite and the `backup-filesystem` module on `ci/docker-compose-rbac.yml` to be dropped ([here](https://github.com/weaviate/typescript-client/pull/459#discussion_r3795609367) and [here](https://github.com/weaviate/typescript-client/pull/459#discussion_r3795586054)), on the grounds that the root-only read-back is server behaviour and not the client's to assert.

So the read-back is no longer covered by a test in either direction. It still works — your 1.38.9 check and mine on 1.38.2 both round-trip — it is just untested here now.

src/collections/backup/types.ts
dudanogueira: @g-despot the four cases are still documented, but compressed to one line, since @bevzzz asked for the comment volume to come down. It now reads: only returned to root users, from `v1.37.6`, and never by `create()` without `waitForCompletion`.

dudanogueira:
Pushed `6395914c`, taking all of @bevzzz's points. The diff is down from **516 insertions to 138**, across 11 files instead of 15. Per-thread replies are inline; the PR description is rewritten.

What came out:

- the client-side base-ID validation, and the ID lowercasing that went with it — the server resolves backup IDs;
- the `in` guards in `parseStatus`;
- `BackupCreateArgs` and `BackupCollectionCreateArgs` — `incrementalBaseBackupId` is now a field on `BackupConfigCreate`;
- all 243 lines of mock tests;
- the root-user suite and the `backup-filesystem` module on `ci/docker-compose-rbac.yml`;
- the comment volume.

Three things I'd like a decision on rather than deciding myself:

1. **Where the argument lives.** Moving it onto `BackupConfigCreate` makes the call `create({ backupId, backend, config: { incrementalBaseBackupId } })`. The Python client has it as a top-level argument next to `config`, matching the wire payload where `incremental_base_backup_id` is a sibling of `config`, not inside it. This is now the one place the two clients disagree on shape. I went with the review; say the word if parity wins.
2. **The version gate is now untested.** It can only fire against a server reporting `<1.37.0`, so no integration test can reach it. I verified both the v3 and v2 paths by hand against 1.36.0. If you'd rather the gate went too, that's another ~30 lines out.
3. **The read-back is now untested in either direction.** Dropping the root-user suite was right on your reasoning, but it does mean nothing here exercises `incrementalBaseBackupId` coming back from `getCreateStatus()` or `list()`. @g-despot verified it on 1.38.9 and I did on 1.38.2.

One fix that fell out of the trim: moving the gate into `BackupCreator.do()` put it inside the v3 `create` try/catch, which was rewrapping it as `WeaviateBackupFailed`. `WeaviateUnsupportedFeatureError` is now rethrown as-is — the gate rejects before anything is sent, so it isn't a backup failure.

@g-despot two of your requests are reverted here; I've flagged both on your original threads so it isn't a silent undo.

dudanogueira:
CI is red, and I've traced it. **None of it is this PR.** Details so anyone can check rather than take my word for it.

### The one that fails on all 7 legs

`test/collections/journey.test.ts > should get the config for the created collection`, on every leg from 1.33 to 1.38.

It is not a version thing — that test does not use the matrix container at all. It connects to the shared **Weaviate Cloud cluster**, so all seven legs hit the same endpoint and fail identically. The single mismatch is the vector index quantizer:

```
- Expected                        + Received
- "quantizer": undefined          + "quantizer": { "bits": 8, "rescoreLimit": 20, "type": "rq" }
```

(The `textAnalyzer: undefined` lines in the same diff are noise — `toEqual` treats undefined-valued keys as absent.)

So the cluster now creates collections with RQ enabled, and the test's expectation predates that. Worth noting this is **not** the OSS default: I ran 1.39.0 locally and a fresh HNSW collection comes back `rq.enabled: false`. It looks specific to how the cloud cluster provisions new collections, and it changed some time after 2026-08-18, which is the last time `main` ran this test green.

Why it cannot be this PR:

- `test/collections/journey.test.ts` is **byte-identical** between this branch and `main` — the branch never touches it.
- The diff is 138 lines, all of it backup code plus the `dbVersionSupport` wiring. It goes nowhere near config parsing, vector index config, or quantizers.
- The client is reporting the quantizer *correctly*; the server really is sending one.

I'd rather not patch it here — it needs a version- or environment-aware expectation, it affects `main` equally, and this PR was just asked to get smaller. Happy to open a separate PR for it if that's useful; say the word.

### The other two

Both are the flakes from my earlier comment, not new:

- `test/users/integration.test.ts` role assign/revoke 404s — 1.37 and 24.x/1.38 legs
- `collections.query > boost > 'timeDecay'` — 22.x/1.38 leg

The identical users pair failed on `main` on 2026-08-13 (run 31690118238) with no backup code in play.

### The backup tests

All green, on every leg. Both new integration tests passed on 24.x/1.38, which is the leg that actually runs them:

```
✓ incremental backups > creates and restores an incremental backup on top of a base backup  10829ms
✓ incremental backups > sends the base backup ID to the server                                  3ms
```

The branch is also ~12 commits behind `main` now. Rebasing won't change any of the above, but I'll do it before merge if you'd like it current.

src/backup/backupCreator.ts
```diff
@@ -90,12 +108,27 @@ export default class BackupCreator extends CommandBase {
...
-      return this._createAndWaitForCompletion(payload);
+    return this.checkIncrementalSupport().then(() =>
+      this.waitForCompletion ? this._createAndWaitForCompletion(payload) : this._create(payload)
+    );
+  };
+
+  /**
+   * Weaviate below v1.37.0 ignores `incremental_base_backup_id` and silently writes a full backup,
+   * so fail loudly instead. No-op without a version provider, or for non-incremental backups.
+   */
```
bevzzz, line 122: nit: the reason for doing "db version check" is well-understood, i.e. to see if this version of DB has this feature, so this comment is likely redundant

src/collections/backup/client.ts
```diff
@@ -47,6 +50,8 @@ export const backup = (connection: Connection): Backup => {
       error: res.error,
       path: res.path,
       status: res.status,
+      size: res.size,
+      incrementalBaseBackupId: res.incremental_base_backup_id || undefined,
```
bevzzz, line 54:
```suggestion
      incrementalBaseBackupId: res.incremental_base_backup_id,
```

nit: this is redundant, `undefined` is already the value that is returned for a missing key

src/collections/backup/client.ts
```diff
@@ -129,6 +137,8 @@ export const backup = (connection: Connection): Backup => {
       try {
         res = await builder.do();
       } catch (err) {
+        // The version gate rejects before anything is sent, so it is not a backup failure.
```
bevzzz, line 140: nit: redundant comment

src/collections/backup/client.ts
```diff
@@ -213,7 +223,12 @@ export const backup = (connection: Connection): Backup => {
       if (opts?.startedAtAsc) {
         url += '?order=asc';
       }
-      return connection.get<BackupReturn[]>(url);
+      return connection.get<BackupListResponse>(url).then((res) =>
+        res.map(({ incremental_base_backup_id: baseBackupId, ...rest }) => ({
+          ...rest,
+          incrementalBaseBackupId: baseBackupId || undefined,
```
bevzzz, line 229:
```suggestion
          incrementalBaseBackupId: baseBackupId,
```

nit: redundant operation

bevzzz (changes requested):

src/collections/backup/client.ts
dudanogueira: Applied in `2e14362d` — and you are right for a reason worth writing down, because the schema comment says the opposite.

`schema.ts` describes the field as "empty if the backup is not incremental", which reads like the server sends `""`. It does not. On the response side the server declares it as a non-pointer string with `omitempty`:

```go
// entities/models/backup_create_status_response.go
IncrementalBaseBackupID string `json:"incremental_base_backup_id,omitempty"`
```

so `""` is dropped before it reaches the wire. Confirmed against the generated models rather than reading the tag:

```
CREATE_STATUS(non-incremental): {"completedAt":"...","id":"b1","startedAt":"..."}
CREATE_STATUS(incremental):     {"completedAt":"...","id":"b2","incremental_base_backup_id":"base1","startedAt":"..."}
```

The key is absent, so the access is already `undefined` and the coalesce never fired. Gone.

src/collections/backup/client.ts
dudanogueira: Applied in `2e14362d`, same reasoning as the `parseStatus` one above — `BackupListResponseItems0.IncrementalBaseBackupID` is also an `omitempty` non-pointer string, so a non-incremental item omits the key rather than sending `""`:

```
LIST_ITEM(non-incremental): {"classes":null,"completedAt":"...","id":"b3","startedAt":"..."}
```

The destructure is now a plain rename.

src/collections/backup/client.ts
dudanogueira: Removed in `2e14362d`.

src/backup/backupCreator.ts
dudanogueira: Trimmed in `2e14362d`. Agreed on the part you named — I had also spelled out that it no-ops without a version provider or for a non-incremental backup, which is just a restatement of the two-line guard directly below it. That sentence is gone.

I kept one line, because the fact it records is the reason the gate exists at all:

```ts
/** Weaviate below v1.37.0 ignores the field and silently writes a full backup, so fail loudly instead. */
```

A version check normally guards a call the server would reject anyway, in which case it is pure ergonomics and needs no comment. Here the server accepts the request and returns a `SUCCESS` that is a *full* backup, so without the gate the failure is silent and only shows up as a surprising restore later. Happy to drop the line too if you would rather have none — say the word and it goes.

PR #460, opened 2026-08-12
Support MMR diversity in Hybrid search

bevzzz:


tsmith023 (approved):

src/collections/query/utils.ts
```diff
@@ -183,4 +183,8 @@ export class Diversity {
       type: 'mmr',
     };
   }
+
+  static isMMR(args: DiversityConfig): args is MMR {
```
g-despot, line 187: There is already a `DiversityGuards.isMMR`, could maybe be reused

PR #462, opened 2026-08-12
Add BM25 AND_CROSS operator

### LINKED ISSUE #458: Add cross-property BM25 And operator
> Mirrored from weaviate/weaviate-python-client#2119 (https://github.com/weaviate/weaviate-python-client/issues/2119)

- Expose the OPERATOR_AND_CROSS search operator (Weaviate 1.39.0, backported to 1.37.15 and 1.38.8).
- Available wherever a BM25 operator is accepted: bm25, hybrid, the generative-query equivalents, and aggregate hybrid.

- Python implementation https://github.com/weaviate/weaviate-python-client/pull/2114

bevzzz:
Closes #458

### DIFF
--- a/src/collections/query/types.ts
+++ b/src/collections/query/types.ts
@@ -93,9 +93,10 @@ export type Bm25QueryProperty<T> = {
 };
 
 export type Bm25OperatorOr = { operator: 'Or'; minimumMatch: number };
-export type Bm25OperatorAnd = { operator: 'And'; crossProperty?: boolean };
+export type Bm25OperatorAnd = { operator: 'And' };
+export type Bm25OperatorAndCross = { operator: 'AndCross' };
 
-export type Bm25OperatorOptions = Bm25OperatorOr | Bm25OperatorAnd;
+export type Bm25OperatorOptions = Bm25OperatorOr | Bm25OperatorAnd | Bm25OperatorAndCross;
 
 export type Bm25SearchOptions<T> = {
   /** Which properties of the collection to perform the keyword search on. */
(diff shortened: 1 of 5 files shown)


src/collections/query/types.ts
```diff
@@ -93,7 +93,7 @@ export type Bm25QueryProperty<T> = {
 };
 
 export type Bm25OperatorOr = { operator: 'Or'; minimumMatch: number };
-export type Bm25OperatorAnd = { operator: 'And' };
+export type Bm25OperatorAnd = { operator: 'And'; crossProperty?: boolean };
```
g-despot: Why not `Bm25Operator.andCross()`just like in Python client?

src/collections/serialize/index.ts
```diff
@@ -1046,7 +1046,9 @@ export class Serialize {
     if (searchOperator) {
       return SearchOperatorOptions.fromPartial(
         searchOperator.operator === ('And' as const)
-          ? { operator: SearchOperatorOptions_Operator.OPERATOR_AND }
+          ? searchOperator.crossProperty
+            ? { operator: SearchOperatorOptions_Operator.OPERATOR_AND_CROSS }
```
g-despot: Do we have a guard anywhere to make sure this value is not sent to a Weaviate server below version 1.39? Seems like it will fail through silently instead of warning the user that it's not supported

tsmith023: I _think_, but would need confirming, that the server throws an error in such a case

src/collections/query/types.ts
bevzzz: I thought it would make sense seeing how `minimumMatch` is an additional parameter to `OR` operator and `crossProperty` can be seen as an additional parameter to `AND` operator.

Changed now.

g-despot (approved):

PR #464, opened 2026-08-30
fix(data): normalize unwrapped NonReferenceInputs in data.ingest to prevent silent data loss

### LINKED ISSUE #456: data.ingest() silently stores empty objects for the NonReferenceInputs form its type signature accepts
## Summary

`collection.data.ingest()` accepts `NonReferenceInputs<T>` (the unwrapped `{ title: 'x' }` form) in its type signature, but silently stores **objects with no properties** when given that form. The call reports success: UUIDs are returned, `errors` is empty, and `hasErrors` is `false`.

`collection.data.insertMany()` handles the same unwrapped input correctly, so the two batch entry points disagree on an input shape they both claim to accept.

This is silent data loss. A caller following the type signature gets a clean success response and an empty collection.

## The type signature accepts the unwrapped form

`src/collections/data/index.ts:88`:

```ts
ingest: (objs: Iterable<DataObject<T> | NonReferenceInputs<T>>) => Promise<BatchObjectsReturn<T>>;
```

## The implementation does not normalize it

`src/collections/data/index.ts:251-257`:

```ts
for (const obj of objs) {
  // eslint-disable-next-line no-await-in-loop
  await batching.addObject({
    collection: name,
    ...obj,          // <-- spread directly
    tenant,
  });
}
```

For a `DataObject<T>` the spread yields `{ collection, properties, tenant }` and works. For a `NonReferenceInputs<T>` it yields `{ collection, title, tenant }` — `properties` is never set, so the object is created empty.

Compare `insert` in the same file, `:287`, which normalizes exactly this case:

```ts
parseObject(
  obj ? (DataGuards.isDataObject(obj) ? obj : ({ properties: obj } as InsertObject<T>)) : obj
),
```

`DataGuards.isDataObject` is defined at `src/collections/serialize/index.ts:355` and is already the established guard for this discrimination. `ingest` never calls it.

## Reproduction

```js
import weaviate from 'weaviate-client';

const client = await weaviate.connectToLocal();

async function fresh(name) {
  if (await client.collections.exists(name)) await client.collections.delete(name);
  return client.collections.create({
    name,
    properties: [{ name: 'title', dataType: 'text' }],
    vectorizers: weaviate.configure.vectors.selfProvided(),
  });
}

// A — unwrapped NonReferenceInputs form
const a = await fresh('ReproIngestUnwrapped');
const ra = await a.data.ingest([{ title: 'alpha' }, { title: 'beta' }]);
console.log('uuids:', Object.keys(ra.uuids).length, 'errors:', Object.keys(ra.errors).length, 'hasErrors:', ra.hasErrors);
for await (const o of a.iterator()) console.log('STORED:', JSON.stringify(o.properties));

// B — wrapped DataObject form
const b = await fresh('ReproIngestWrapped');
await b.data.ingest([{ properties: { title: 'alpha' } }, { properties: { title: 'beta' } }]);
for await (const o of b.iterator()) console.log('STORED:', JSON.stringify(o.properties));

// C — insertMany with the same unwrapped form
const c = await fresh('ReproInsertMany');
await c.data.insertMany([{ title: 'alpha' }, { title: 'beta' }]);
for await (const o of c.iterator()) console.log('STORED:', JSON.stringify(o.properties));
```

### Actual output

```
A: data.ingest([{ title: "..." }])          <- NonReferenceInputs form
  reported uuids   : 2
  reported errors  : 0
  hasErrors        : false
  STORED PROPERTIES: {}
  STORED PROPERTIES: {}

B: data.ingest([{ properties: { title } }]) <- DataObject form
  reported uuids   : 2
  STORED PROPERTIES: {"title":"beta"}
  STORED PROPERTIES: {"title":"alpha"}

C: data.insertMany([{ title: "..." }])      <- same unwrapped form
  STORED PROPERTIES: {"title":"alpha"}
  STORED PROPERTIES: {"title":"beta"}
```

### Expected

Case A should store `{"title":"alpha"}` and `{"title":"beta"}`, matching cases B and C.

If the unwrapped form is not intended to be supported by `ingest`, then the type signature at `:88` should drop `NonReferenceInputs<T>` so this fails at compile time rather than silently at runtime.

## Suggested fix

Apply the same normalization `insert` already uses, inside the `ingest` loop:

```ts
for (const obj of objs) {
  await batching.addObject({
    collection: name,
    ...(DataGuards.isDataObject(obj) ? obj : { properties: obj }),
    tenant,
  });
}
```

## Environment

- `weaviate-client` 3.13.1
- Weaviate server 1.38.0 (`cr.weaviate.io/semitechnologies/weaviate:1.38.0`, anonymous access, self-provided vectors)
- Node.js v22.14.0, macOS

## Note

This surfaced while updating the Weaviate quickstart documentation to use `data.ingest()`. Every plain property list had to be rewritten as `.map((properties) => ({ properties }))` to work, which is what led back to the signature mismatch.

OsamaAnsar:
## Summary

Fixes #456.

`collection.data.ingest()` accepts `NonReferenceInputs<T>` (the unwrapped `{ title: 'x' }` form) per its own type signature, but the implementation spread the object directly into the `BatchObject` payload without normalizing it into a `properties` field first. `insert()` already does this correctly via `DataGuards.isDataObject(...) ? obj : { properties: obj }` — `ingest()` was the one call site that skipped it, so a caller following the documented type signature got a clean success response (valid UUIDs, no errors) while the object was actually stored empty.

## Fix

Extracted the same normalization `insert()` already uses into a small exported helper, `normalizeBatchInput`, and used it in the `ingest()` loop instead of the raw spread.

## Testing

Added 5 unit tests covering:
- an unwrapped `NonReferenceInputs` object gets wrapped into `properties`
- an already-wrapped `DataObject` (by `id`, `properties`, `references`, or `vectors`) is left untouched
- the unwrapped and wrapped forms produce identical `properties` after normalization

All existing unit tests continue to pass. I wasn't able to run the `integration`/`journey` test suites locally (they require a live Weaviate server via `docker-compose`, not available in this environment) — happy to have those validated in CI, or to add an integration test case mirroring the issue's repro script if useful.

### DIFF
--- a/src/collections/data/index.ts
+++ b/src/collections/data/index.ts
@@ -180,6 +180,14 @@ type ParseObject<T> = {
   vectors?: number[] | Vectors;
 };
 
+/**
+ * Normalizes an object passed to `data.ingest` / `data.insertMany` into the wrapped
+ * `DataObject` shape. Without this, an unwrapped `NonReferenceInputs` value (e.g. `{ title: 'x' }`)
+ * has no `properties` key and is silently stored as an empty object.
+ */
+export const normalizeBatchInput = <T>(obj: DataObject<T> | NonReferenceInputs<T>): DataObject<T> =>
+  DataGuards.isDataObject(obj) ? obj : ({ properties: obj } as DataObject<T>);
+
 const data = <T>(
   connection: Connection,
   name: string,
@@ -252,7 +260,7 @@ const data = <T>(
         // eslint-disable-next-line no-await-in-loop
         await batching.addObject({
           collection: name,
-          ...obj,
+          ...normalizeBatchInput(obj),
           tenant,
         });
       }
--- a/src/collections/data/unit.test.ts
+++ b/src/collections/data/unit.test.ts
@@ -1,6 +1,7 @@
 /* eslint-disable no-await-in-loop */
 import { describe, expect, it } from 'vitest';
 import { Queue } from './batch';
+import { normalizeBatchInput } from './index';
 
 describe('Unit testing of the Queue class', () => {
   it('should push and pull items in FIFO order', async () => {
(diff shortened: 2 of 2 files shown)


OsamaAnsar:
agree with the CLA

PR #466, opened 2026-08-31
feat(vectorizer): add endpoint setting to text2vec-openai and text2vec-morph

### LINKED ISSUE #444: Add support for `endpoint` setting in `text2vec-openai` and `text2vec-morph` modules
> Mirrored from weaviate/weaviate-python-client#2072 (https://github.com/weaviate/weaviate-python-client/issues/2072)

Add support for `endpoint` setting in `text2vec-openai` and `text2vec-morph` modules.

This is an optional setting which defaults to `/v1/embeddings` in Weaviate.

OsamaAnsar:
Closes #444

Adds an optional `endpoint` field to `Text2VecOpenAIConfig` and `Text2VecMorphConfig`, matching what the issue requested — an override for the embeddings endpoint path, defaulting to `/v1/embeddings` in Weaviate when not set.

## Why this was low-risk

Both vectorizer builders (`text2VecOpenAI`, `text2VecMorph`) already spread all config fields through as-is (`config: Object.keys(config).length === 0 ? undefined : config`), so no builder logic needed to change — this is purely additive to the two config type definitions.

## Testing

- Extended the existing "with all values" unit tests for both configs to cover `endpoint`.
- `tsc --noEmit`: clean.
- `vitest run src/collections/configure/unit.test.ts`: 136/136 passing.

### DIFF
--- a/src/collections/config/types/vectorizer.ts
+++ b/src/collections/config/types/vectorizer.ts
@@ -587,6 +587,8 @@ export type Text2VecOpenAIConfig = {
   baseURL?: string;
   /** The dimensionality of the vector once embedded. */
   dimensions?: number;
+  /** The endpoint path to use for embedding requests. Defaults to `/v1/embeddings` in Weaviate. */
+  endpoint?: string;
   /** The model to use. */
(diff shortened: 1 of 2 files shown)


OsamaAnsar:
agree with the CLA

Issue #470, opened 2026-09-01
Flaky CI: intermittent tests-without-auth failures (boost timeDecay ~1-in-6, plus two confirmed serialization bugs)


dudanogueira:
## Summary

`tests-without-auth` fails intermittently across all Weaviate versions and both Node versions, on unrelated PRs. Re-running the identical commit turns it green, so the code under test is not the cause.

Most recent evidence: [run 33513846023](https://github.com/dudanogueira/typescript-client/actions/runs/33513846023) failed, then passed on all 8 jobs after a re-run with no code change.

At least three distinct flakes are involved. One is now understood in detail; the other two are not.

---

## 1. `query > boost > 'timeDecay'` — understood, ~1-in-6 failure rate

`test/collections/query/integration.test.ts`

```
AssertionError: expected [ …(3) ] to not deeply equal [ …(3) ]
```

The assertion is:

```ts
const baseline = ids(await collection.query.nearVector([1, 2, 3]));
const boosted  = ids(await collection.query.nearVector([1, 2, 3], { boost }));
expect(boosted).not.toEqual(baseline);
```

### Measured behaviour

Reproduced locally against `semitechnologies/weaviate:1.38.2`, rebuilding the collection each trial:

- **5/30 trials (16.7%) failed** on unmodified `main`.
- Across rebuilds the boosted ordering spans **all six permutations** of the three objects, near-uniformly. 1/6 = 16.7%, matching the observed rate.
- Within a *single* collection, both baseline and boosted are perfectly stable (0/30 unstable). This is why re-running the same query never reproduces it — only rebuilding the collection does.

### Ruled out

| Hypothesis | Result |
|---|---|
| All three `createdAt` set to the same `Date.now()` | ruled out — spreading them failed at the same rate |
| `blue`/`green` share vector `[1,2,3]` (distance tie) | ruled out — distinct vectors failed at the same rate |
| Query racing vector-index build after `insertMany` | ruled out — 250ms and 1000ms waits changed nothing |

### Assessment

The test's premise is the problem: it assumes a boost *must* reorder three near-tied results, which the server does not guarantee. Retuning the input values will not hold — 467e5a7b already attempted exactly that in July, and this is the same test failing again.

Suggested direction: assert something deterministic (an expected ordering over well-separated data, or scores/membership) rather than "the order changed". The other three boost cases in the same `it.each` (`filter`, `numericDecay`, `numericProperty`) have never flaked.

---

[...]

PR #471, opened 2026-09-02
fix(query): map Boost decay field to proto decayValue, remove debug log

OsamaAnsar:
## Summary

Addresses two of the confirmed bugs from #470 (items 2a and 2b), plus a small related test hardening.

`Serialize.boostGRPC` spread the public `decay` field (on both `TimeDecay` and `NumericDecay`) directly into `Boost_TimeDecayFunction.fromPartial({...})` / `Boost_NumericDecayFunction.fromPartial({...})`. The generated proto type only recognizes `decayValue`, not `decay`, so `fromPartial` silently dropped the field — any caller passing `decay` today has it ignored, with no error, and the value never reaches the wire.

Separately, a `console.error(JSON.stringify(out))` debug line was left in the same function, writing to stderr on every boosted query.

## Fix

- Destructure `decay` alongside `curve` in both the `timeDecay` and `numericDecay` branches and map it explicitly to `decayValue` in the `fromPartial({...})` call.
- Remove the stray `console.error`.
- `test/collections/journey.test.ts`'s `afterAll` called `client.collections.delete(...)` unguarded; if `beforeAll` fails, `client` is `undefined` and this throws a second, misleading `TypeError` that obscures the real failure. Guarded with `client?.`.

## Testing

Added 3 unit tests in `src/collections/serialize/unit.test.ts` (pure serializer, no live server required):
- `TimeDecay.decay` maps onto `decayValue`
- `NumericDecay.decay` maps onto `decayValue`
- `decayValue` is omitted when `decay` isn't provided

Full serialize suite (64 tests) passes, `tsc --noEmit` is clean, and `eslint` on the changed files is clean.

I did not touch the flaky `query > boost > 'timeDecay'` ordering test (item 1 in #470) or the UUID stringify bug on `dryRun deleteMany` (item 3a) — both need more investigation/judgment than this PR's scope, and #470 itself marks them as not yet understood.

### DIFF
--- a/src/collections/serialize/unit.test.ts
+++ b/src/collections/serialize/unit.test.ts
@@ -1034,7 +1034,7 @@ describe('Unit testing of Serialize', () => {
   });
 
   describe('.boostGRPC', () => {
-    it('should map TimeDecay.decay onto the proto decayValue field', () => {
+    it('should serialize a TimeDecay boost to the full proto shape with decay mapped to decayValue', () => {
       const out = Serialize.boostGRPC({
         conditions: [
           {
@@ -1084,21 +1084,6 @@ describe('Unit testing of Serialize', () => {
         ],
       });
       expect(out.conditions[0].numericDecay?.decayValue).toEqual(0.5);
-    });
-
-    it('should omit decayValue when decay is not provided', () => {
-      const out = Serialize.boostGRPC({
-        conditions: [
-          {
-            func: {
-              type: 'timeDecay',
-              property: 'createdAt',
-              scale: '35d',
-            },
-          },
-        ],
-      });
-      expect(out.conditions[0].timeDecay?.decayValue).toBeUndefined();
     });
   });
 });
--- a/src/collections/serialize/unit.test.ts
+++ b/src/collections/serialize/unit.test.ts
@@ -34,7 +34,7 @@ import {
   Targets,
 } from '../../proto/v1/base_search.js';
 import { GenerativeSearch } from '../../proto/v1/generative.js';
-import { GroupBy, MetadataRequest, PropertiesRequest } from '../../proto/v1/search_get.js';
+import { Boost_DecayCurve, GroupBy, MetadataRequest, PropertiesRequest } from '../../proto/v1/search_get.js';
 import { Filters as FiltersFactory } from '../filters/classes.js';
 import filter from '../filters/index.js';
 import { Bm25OperatorOptions, TargetVectorInputType } from '../query/types.js';
@@ -1034,9 +1034,6 @@ describe('Unit testing of Serialize', () => {
   });
 
   describe('.boostGRPC', () => {
-    // Regression coverage: `decay` was being spread into `fromPartial({...})` unchanged,
-    // but the generated proto type only recognizes `decayValue`, so `fromPartial` silently
-    // dropped it and the value never reached the wire. See weaviate/typescript-client#470.
     it('should map TimeDecay.decay onto the proto decayValue field', () => {
       const out = Serialize.boostGRPC({
         conditions: [
@@ -1052,7 +1049,23 @@ describe('Unit testing of Serialize', () => {
           },
         ],
       });
-      expect(out.conditions[0].timeDecay?.decayValue).toEqual(0.2);
+      expect(out).toEqual({
+        weight: undefined,
+        depth: undefined,
+        conditions: [
+          {
+            weight: undefined,
+            timeDecay: {
+              property: 'createdAt',
+              origin: '2024-01-01T00:00:00Z',
+              scale: '35d',
+              offset: undefined,
+              curve: Boost_DecayCurve.DECAY_CURVE_EXPONENTIAL,
+              decayValue: 0.2,
+            },
+          },
+        ],
+      });
     });
 
     it('should map NumericDecay.decay onto the proto decayValue field', () => {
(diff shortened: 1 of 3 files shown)


src/collections/serialize/unit.test.ts
```diff
@@ -1032,6 +1032,62 @@ describe('Unit testing of Serialize', () => {
       });
     });
   });
+
+  describe('.boostGRPC', () => {
+    // Regression coverage: `decay` was being spread into `fromPartial({...})` unchanged,
+    // but the generated proto type only recognizes `decayValue`, so `fromPartial` silently
+    // dropped it and the value never reached the wire. See weaviate/typescript-client#470.
```
bevzzz: nit: please remove this comment, it isn't necessary

src/collections/serialize/unit.test.ts
```diff
@@ -1032,6 +1032,62 @@ describe('Unit testing of Serialize', () => {
...
+              scale: '35d',
+              origin: '2024-01-01T00:00:00Z',
+              curve: 'exponential',
+              decay: 0.2,
+            },
+          },
+        ],
+      });
+      expect(out.conditions[0].timeDecay?.decayValue).toEqual(0.2);
+    });
```
bevzzz: suggestion: I think we can extend this test to compare the entire output object and check that it matches the expected proto stub.

OsamaAnsar:
I agree with the CLA

OsamaAnsar:
Thanks for the review! Both addressed in 7bb9971: removed the comment, and extended the TimeDecay test to assert the full `boostGRPC` output against the expected proto stub instead of just `decayValue`.

src/collections/serialize/unit.test.ts
```diff
@@ -1032,6 +1032,75 @@ describe('Unit testing of Serialize', () => {
       });
     });
   });
+
+  describe('.boostGRPC', () => {
+    it('should map TimeDecay.decay onto the proto decayValue field', () => {
```
bevzzz: nit: name of the test should be updated too, now that the entire proto value is compared

src/collections/serialize/unit.test.ts
```diff
@@ -1032,6 +1032,75 @@ describe('Unit testing of Serialize', () => {
...
+            func: {
+              type: 'timeDecay',
+              property: 'createdAt',
+              scale: '35d',
+            },
+          },
+        ],
+      });
+      expect(out.conditions[0].timeDecay?.decayValue).toBeUndefined();
+    });
```
bevzzz: nit: IMO this test is redundant, there's nothing in the codebase to suggest an "inexistent" `decayValue` may be passed

bevzzz (changes requested):

OsamaAnsar:
Thanks @bevzzz — both addressed in c502ab7:

- Renamed the test to reflect that it now asserts the full proto `Boost` shape, not just the mapped field.
- Dropped the `should omit decayValue when decay is not provided` test as redundant.

Serialize unit suite green (63 passing), lint / typecheck / prettier clean.

src/collections/serialize/unit.test.ts
```diff
@@ -1032,6 +1032,60 @@ describe('Unit testing of Serialize', () => {
...
+              offset: undefined,
+              curve: Boost_DecayCurve.DECAY_CURVE_EXPONENTIAL,
+              decayValue: 0.2,
+            },
+          },
+        ],
+      });
+    });
+
+    it('should map NumericDecay.decay onto the proto decayValue field', () => {
```
bevzzz: [This](https://github.com/weaviate/typescript-client/pull/471#discussion_r3923208071) is also applicable here ( the test should check the entire proto structure )

bevzzz (changes requested):

bevzzz:
@OsamaAnsar please take care to trim PR descriptions to what is strictly relevant in the future.

OsamaAnsar:
> @OsamaAnsar please take care to trim PR descriptions to what is strictly relevant in the future.

sure I will take care of this in future

OsamaAnsar:
Addressed in 3dee784 — the `NumericDecay` test now asserts the entire proto output via `toEqual`, matching the `TimeDecay` test, and is renamed accordingly. Serialize unit suite green (63).

test/collections/journey.test.ts
```diff
@@ -13,7 +13,7 @@ describe('Journey testing of the client using a WCD cluster', () => {
     dateOfBirth: Date;
   };
 
-  afterAll(() => client.collections.delete(collectionName));
+  afterAll(() => client?.collections.delete(collectionName));
```
bevzzz: question: What motivates this change? We don't do `client?.` anywhere else in the this test suite.

OsamaAnsar:
Good question — dropped it. That `client?.` and the quantizer-assertion edits were stale: this branch predated #469, which already makes the journey suite tolerant of the shared cluster's default quantizer. Rebased onto `main` and reverted `journey.test.ts` entirely, so #471 is now just the Boost-decay serialize fix (`serialize/index.ts` + `serialize/unit.test.ts`). Serialize unit suite green (63).

OsamaAnsar:
@bevzzz all six points from the review are addressed in the latest push (047028b):

- dropped the regression comment
- both `boostGRPC` tests now assert the full proto object via `toEqual`, and are renamed accordingly
- removed the redundant "omitted `decayValue`" test
- reverted the `client?.` guard in `journey.test.ts` back to `client.`

`test:unit` (serialize suite, 63 tests), `tsc --noEmit`, and eslint on the changed files are all green locally. The remaining red check is the unrelated `query > boost > 'timeDecay'` ordering flake (item 1 in #470), untouched here. Could you take another look when you have a moment?

bevzzz (approved):

PR #476, opened 2026-09-07
feat(generative): add location setting to generative-google config

### LINKED ISSUE #445: Add support for `location` setting in `generative-google` module
> Mirrored from weaviate/weaviate-python-client#2073 (https://github.com/weaviate/weaviate-python-client/issues/2073)

Add support for `location` setting in `generative-google` module

This is an optional setting which defaults to `us-central1` in Weaviate.

This change needs newest protos in order for the clients to support this new setting

OsamaAnsar:
## Summary

Closes #445.

Adds the optional `location` setting to the `generative-google` module (defaults to `us-central1` server-side). Mirrors weaviate-python-client#2118 / #2130.

## Approach

`location` is a plain pass-through field on both existing `generative-google` config surfaces, so no builder logic changes are needed:

- **Config time** — `configure.generative.google()` already spreads `...config`, so adding `location` to `GenerativeGoogleConfig` is enough.
- **Query time** — `generativeParameters.google()` already spreads `...rest`, and the vendored `GenerativeGoogle` proto message already carries `location` (`src/proto/v1/generative.ts`), so adding `location` to `GenerativeGoogleConfigRuntime` completes the runtime path without touching `serialize/index.ts`.

## Changes

- `src/collections/config/types/generative.ts` — `location?: string` on `GenerativeGoogleConfig`
- `src/collections/types/generate.ts` — `location?: string` on `GenerativeGoogleConfigRuntime`
- `src/collections/configure/unit.test.ts` — `location` added to the existing `generative.google` all-values test

No new test in `generate/unit.test.ts`: per that file's header comment it only covers fields that are renamed to a gRPC name, and `location` passes through unchanged.

## Testing

`npx tsc --noEmit` clean. `eslint` and `prettier --check` clean. Unit suite passes (160/160 across `configure/unit.test.ts` + `generate/unit.test.ts`) with `WEAVIATE_VERSION` set locally for `test/version.ts`.

### DIFF
--- a/src/collections/config/types/generative.ts
+++ b/src/collections/config/types/generative.ts
@@ -88,6 +88,7 @@ export type GenerativePaLMConfig = GenerativeGoogleConfig;
 
 export type GenerativeGoogleConfig = {
   apiEndpoint?: string;
+  location?: string;
   maxOutputTokens?: number;
   model?: string;
   /** @deprecated Use `model` instead. */
--- a/src/collections/configure/unit.test.ts
+++ b/src/collections/configure/unit.test.ts
@@ -2226,6 +2226,7 @@ describe('Unit testing of the generative factory class', () => {
   it('should create the correct GenerativeGoogleConfig type with all values', () => {
     const config = configure.generative.google({
       apiEndpoint: 'api-endpoint',
+      location: 'europe-west4',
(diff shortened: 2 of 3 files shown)


OsamaAnsar:
I agree with the CLA

